PR #10, opened 2019-05-22
Add EventType enum to Event

NickGrantONS:
# Motivation and Context
We need an event type stored against the case events to make it easier to find specific ones. The event types should come from a discrete enumerated list.

# What has changed
Added `EventType` enum and added it as a column to the `Event` entity.

# How to test?
Run ATs and check `event_type` DB column is populated in the `event` table in the case DB.

# Links
Trello: https://trello.com/c/ECRAB4bx


NickGrantONS:
[Add "event type" to case events (5)](https://trello.com/c/ECRAB4bx/748-add-event-type-to-case-events-5)

PR #21, opened 2019-06-04
RMC-152

rjweeks:
# Motivation and Context
We need to record when a reciept has been received from EQ and issue a Case Updated event,

# What has changed
UAC active and receipt received booleans added to Case Processor DB.

# How to test?
Build local images of Case Processor, Case API and Action Scheduler.
Run acceptance tests.

# Links
https://trello.com/c/GdA7flOq/788-rmc-152a-receipting-should-update-case-and-uac-active-flag-8


AdeHarrison:
[RMC-152(A) Receipting should update case and UAC active flag (8)](https://trello.com/c/GdA7flOq/788-rmc-152a-receipting-should-update-case-and-uac-active-flag-8)

PR #22, opened 2019-06-21
Add specific routing keys for case.action to replace wildcard key

hbrace:
# Motivation and Context

Currently the case.action queue has a wildcard as the routing key. This is causing the action-scheduler issues as it's trying to consume messages it can't.

# What has changed

Two routing keys are now bound to the action.case queue.

# How to test?

Build image off this branch, push to your GCR and deploy by tweaking [this line](https://github.com/ONSdigital/census-rm-kubernetes/blob/master/microservices/action-scheduler-deployment.yml#L37) to point at your pushed image. Check in rabbitmq management that the action.case queue has the following bindings:
image

**If you do not stop and start rabbitmq the old wildcard routing key will still be present, press the unbind button to get rid of it. This will also need to be done if deploying to any environment where rabbitmq is not restarted when deploying this change.**

Run the acceptance tests, also you can try publishing a message to the exchange from rabbitmq management with a routing key that does not exist and checking that the pop up says "Message published, but not routed."

# Links

https://trello.com/c/klIhVlAW/871-change-rabbit-topic-binding-for-caseaction-queue-to-be-specific-to-case-uac-events

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/config/QueueSetterUpper.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/config/QueueSetterUpper.java
@@ -33,8 +33,11 @@ public class QueueSetterUpper {
   @Value("${queueconfig.action-scheduler-queue}")
   private String actionSchedulerQueue;
 
-  @Value("${queueconfig.action-scheduler-routing-key}")
-  private String actionSchedulerRoutingKey;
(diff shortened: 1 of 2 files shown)


hbrace:
[change rabbit topic binding for case.action queue to be specific to case/uac events (3?)](https://trello.com/c/klIhVlAW/871-change-rabbit-topic-binding-for-caseaction-queue-to-be-specific-to-case-uac-events)

PR #25, opened 2019-07-03
Make caseRef randomy generated

NickGrantONS:
# Motivation and Context

Case ref needs to be random rather than sequential because reasons.

# What has changed

Got rid of sequence generator and replaced with random number generator instead.

# How to test?

Load a sample. Check that all the case refs are random.

# Links

Trello: https://trello.com/c/ky7qtJCe

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/utility/RandomCaseRefGenerator.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/utility/RandomCaseRefGenerator.java
@@ -8,7 +8,7 @@ public class RandomCaseRefGenerator {
   private static final int HIGHEST_POSSIBLE_CASE_REF = 99999999;
   private static final Random random = new SecureRandom();
 
-  public static int testGetCaseRef() {
+  public static int getCaseRef() {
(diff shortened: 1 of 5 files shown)


NickGrantONS:
[caseref has to be random ** (5)](https://trello.com/c/ky7qtJCe/798-caseref-has-to-be-random-5)

src/main/java/uk/gov/ons/census/casesvc/utility/RandomCaseRefGenerator.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.security.SecureRandom;
+import java.util.Random;
+
+public class RandomCaseRefGenerator {
+  private static final int LOWEST_POSSIBLE_CASE_REF = 10000000;
+  private static final int HIGHEST_POSSIBLE_CASE_REF = 99999999;
+  private static final Random random = new SecureRandom();
+
+  public static int testGetCaseRef() {
```
liamtoozer: Does this need to be `getCaseRef()` instead?

NickGrantONS: fixed

PR #28, opened 2019-07-16
Check for case ref collisons

NickGrantONS:
# Motivation and Context
In the event of a case ref collision Hibernate is not throwing an ID not unique exception, but instead updating an existing entity, causing us to overwrite case data. We need to check that our ID is unique before we proceed with it and we need to flush it to the DB so that we guarantee that new cases are non-colliding.

# What has changed
Check for uniqueness of caseref. Use 'saveAndFlush' instead of 'save'.

# How to test?
Fill up the `cases` table with all possible caserefs using the SQL `INSERT INTO casev2.cases (case_ref) SELECT * FROM generate_series(10000000, 99999999);` and then ensure that the cases are valid using the SQL `update casev2.cases set receipt_received='f';`. Then try loading a sample - the messages should never be consumed from the `case.sample.inbound` queue and no messages should be published.

# Links
Trello: https://trello.com/c/seupU0RG

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/messaging/SampleReceiver.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/messaging/SampleReceiver.java
@@ -1,6 +1,5 @@
 package uk.gov.ons.census.casesvc.messaging;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import org.springframework.integration.annotation.MessageEndpoint;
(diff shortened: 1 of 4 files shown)


NickGrantONS:
[DEFECT: Colliding caserefs are not causing Rabbit transaction rollback](https://trello.com/c/seupU0RG/963-defect-colliding-caserefs-are-not-causing-rabbit-transaction-rollback)

PR #32, opened 2019-08-01
Dates serialised in UTC format

AdeHarrison:
# Motivation and Context
Dates are not serialised to JSON in UTC format

# What has changed
Changed OffsetDateTime field type to String

# How to test?
build local image of this branch and run with docker dev
run acceptance tests and check date formats in event_payload field in events table

# Links
[Trello](https://trello.com/c/eF4c7Q7e/979-dates-in-case-event-payload-json-not-being-stored-correctly-3)

### DIFF
--- a/src/test/java/uk/gov/ons/census/casesvc/service/RefusalProcessorTest.java
+++ b/src/test/java/uk/gov/ons/census/casesvc/service/RefusalProcessorTest.java
@@ -43,14 +43,15 @@ public class RefusalProcessorTest {
   public void shouldProcessARefusalReceivedMessageSuccessfully() {
     // GIVEN
     ResponseManagementEvent managementEvent = getTestResponseManagementEvent();
-    managementEvent.getPayload().getRefusal().setResponseDateTime(OffsetDateTime.now().toString());
+    managementEvent.getPayload().getRefusal().setResponseDateTime(OffsetDateTime.now());
     Case testCase = getRandomCase();
     testCase.setRefusalReceived(false);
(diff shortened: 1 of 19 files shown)


AdeHarrison:
[dates in case event payload json not being stored correctly (3)](https://trello.com/c/eF4c7Q7e/979-dates-in-case-event-payload-json-not-being-stored-correctly-3)

src/test/java/uk/gov/ons/census/casesvc/service/RefusalProcessorTest.java
```diff
@@ -71,10 +70,10 @@ public void shouldProcessARefusalReceivedMessageSuccessfully() {
             EventType.CASE_UPDATED,
             expectedRefusal,
             managementEvent.getEvent(),
-            expectedRefusal.getResponseDateTime());
+            OffsetDateTime.parse(expectedRefusal.getResponseDateTime()));
   }
 
-  @Test(expected = RuntimeException.class)
```
AdamHawtin: Without re-throwing and expecting the exception here this test will pass even if the exception is not thrown

AdeHarrison: not sure what you mean here @AdamHawtin, test fails unless RuntimeException?

AdamHawtin: If a runtime exception is thrown and the message doesn't match, it'll fail but if not exception is thrown the test will still pass.
You can comment out `underTest.processRefusal(managementEvent);` and the test passes.

PR #39, opened 2019-08-13
persist deactivation of uac/qid to database

AdeHarrison:
# Motivation and Context
When a case receipt notification is received, the uac/qid link active state property is not persisted.

# What has changed
uac/qid link active state persisted to database

# How to test?
build local docker image from branch
run with docker dev
run tests

# Links
[Trello](https://trello.com/c/PVJDUlm5/985-deactivate-uac-qid-on-receipt-of-responsereceived-event-3)

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/ReceiptProcessor.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/ReceiptProcessor.java
@@ -58,12 +58,16 @@ public class ReceiptProcessor {
     }
 
     UacQidLink uacQidLink = uacQidLinkOpt.get();
-    Case caze = uacQidLink.getCaze();
+    uacQidLink.setActive(false);
+    uacQidLinkRepository.saveAndFlush(uacQidLink);
 
-    uacProcessor.emitUacUpdatedEvent(uacQidLink, caze, false);
+    Case caze = uacQidLink.getCaze();
     caze.setReceiptReceived(true);
     caseRepository.saveAndFlush(caze);
+
+    uacProcessor.emitUacUpdatedEvent(uacQidLink, caze, uacQidLink.isActive());
     caseProcessor.emitCaseUpdatedEvent(caze);
+
     eventLogger.logReceiptEvent(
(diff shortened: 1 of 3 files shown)


AdeHarrison:
[Deactivate UAC/QID on receipt of ResponseReceived event (3)](https://trello.com/c/PVJDUlm5/985-deactivate-uac-qid-on-receipt-of-responsereceived-event-3)

src/main/java/uk/gov/ons/census/casesvc/service/ReceiptProcessor.java
```diff
@@ -58,12 +58,16 @@ public void processReceipt(ResponseManagementEvent receiptEvent) {
     }
 
     UacQidLink uacQidLink = uacQidLinkOpt.get();
-    Case caze = uacQidLink.getCaze();
+    uacQidLink.setActive(false);
```
rjweeks: This could be called receiptedCase instead of "caze"

AdeHarrison: good ideas @rjweeks but it's caze everywhere else so reluctant to change it

src/main/java/uk/gov/ons/census/casesvc/service/ReceiptProcessor.java
```diff
@@ -58,12 +58,16 @@ public void processReceipt(ResponseManagementEvent receiptEvent) {
...
+    uacQidLinkRepository.saveAndFlush(uacQidLink);
 
-    uacProcessor.emitUacUpdatedEvent(uacQidLink, caze, false);
+    Case caze = uacQidLink.getCaze();
     caze.setReceiptReceived(true);
     caseRepository.saveAndFlush(caze);
+
+    uacProcessor.emitUacUpdatedEvent(uacQidLink, caze, uacQidLink.isActive());
```
NickGrantONS: why do you need to pass in the isActive if you're already passing in the uacQidLink (which contains the isActive)?

AdeHarrison: because basically, it's a one line fix, didn't want to start refactoring to that level - emitUacUpdatedEvent() is overloaded which always passes in "true" - needs a cleanup but not now

PR #42, opened 2019-08-20
Handle individual response fulfilment requests

liamtoozer:
# Motivation and Context

We need to handle Individual response (IR) household fulfilment requests.

# What has changed

* Validate the fulfilment code
* Generate and save a new IR case with an address type of `HI`
* Emit a `CASE_CREATED` event for the new IR case
* Moved validation logic to its own function for fulfilment events
* Added tests

# How to test?

Run ATs using branch from [this related PR](https://github.com/ONSdigital/census-rm-acceptance-tests/pull/109)

# Links

https://trello.com/c/xem9Ksud/878-rmc-211a-handle-request-for-individual-uac-8


liamtoozer:
[RMC-211(A) Handle request for Individual UAC (8)](https://trello.com/c/xem9Ksud/878-rmc-211a-handle-request-for-individual-uac-8)

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,100 @@
   private static final String CASE_NOT_FOUND_ERROR = "Case not found error";
   private static final String DATETIME_NOT_PRESENT = "Date time not in event error";
   private static final String FULFILMENT_REQUEST_RECEIVED = "Fulfilment Request Received";
+  private static final String FULFILMENT_CODE_NOT_FOUND = "Fulfilment Code not found";
+
+  private static final Set<String> individualResponseRequestCodes =
+      new HashSet<>(Arrays.asList("UACIT1", "UACIT2", "UACIT2W", "UACIT4"));
```
LukeLoze1: These could be fully explained in Constants and then inserted here.  Also have constants in the tests.  eg.   private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_ENGLAND  = "UACIT1".  I know it's a mouth full, but it also makes it clear in 6 months time when you're wondering what it is.

src/test/java/uk/gov/ons/census/casesvc/messaging/FulfilmentRequestReceiverIT.java
```diff
@@ -90,6 +100,70 @@ public void testFulfilmentRequestLogged() throws InterruptedException {
...
+    // WHEN
+    String json = convertObjectToJson(managementEvent);
+    Message message =
+        MessageBuilder.withBody(json.getBytes())
+            .setContentType(MessageProperties.CONTENT_TYPE_JSON)
+            .build();
+    rabbitQueueHelper.sendMessage(inboundQueue, message);
+
+    // wait for the the emitted event
+    BlockingQueue<String> outboundQueue = rabbitQueueHelper.listen(rhCaseQueue);
```
LukeLoze1: BlockingQueue should be setup before the msg is sent to caseprocessor.  I think this happens to work because the queue gets 'blocked' before case processor can process the msg sent to it,  it's possible though unlikely the case will be emitted before we've blocked the queue.  I think most tests like this declare the blocking queues at the top

src/test/java/uk/gov/ons/census/casesvc/messaging/FulfilmentRequestReceiverIT.java
```diff
@@ -90,6 +100,70 @@ public void testFulfilmentRequestLogged() throws InterruptedException {
...
+    rabbitQueueHelper.checkExpectedMessageReceived(outboundQueue);
```
LukeLoze1: This checks that a msg was sent, it would also be good to check the msg a little: 

copied from SampleReceiverIT:

ResponseManagementEvent responseManagementEvent =
        rabbitQueueHelper.checkExpectedMessageReceived(queue);
    assertEquals(EventType.CASE_CREATED, responseManagementEvent.getEvent().getType());

Maybe also test the EstabArid is there.  responseManagementEvent.payload.collectionCase.address.estabArid . == parent.estabArid

Also wouldn't hurt to stash this as an emittedCase and then compare Id to the saved one later, I know they'll match, but potentially nice

src/test/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessorTest.java
```diff
@@ -55,6 +68,59 @@ public void testGoodFulfilmentRequest() {
             FULFILMENT_REQUESTED,
             expectedFulfilmentRequest,
             managementEvent.getEvent());
+
+    verify(caseRepository, never()).save(any(Case.class));
+    verifyNoMoreInteractions(eventLogger);
+    verifyZeroInteractions(caseProcessor);
```
LukeLoze1: nice

src/test/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessorTest.java
```diff
@@ -103,4 +169,89 @@ public void testNullDateTime() {
...
+            "Fulfilment Request Received",
+            FULFILMENT_REQUESTED,
+            expectedFulfilmentRequest,
+            managementEvent.getEvent());
+
+    ArgumentCaptor<Case> caseArgumentCaptor = ArgumentCaptor.forClass(Case.class);
+    verify(caseRepository).save(caseArgumentCaptor.capture());
+    Case actualChildCase = caseArgumentCaptor.getValue();
+
+    testIndivdualFulfilmentRequestCase(parentCase, actualChildCase);
```
LukeLoze1: probably rename to checkIndivdualFulfilmentRequestCase, as it's not a test per se etc and so on

src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
```diff
@@ -10,10 +10,16 @@
...
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 @Data
 @Entity
+@Builder(toBuilder = true)
```
NickGrantONS: why all the extra annotations? We're going to drown in annotation hell if we annotate everything like this

src/main/java/uk/gov/ons/census/casesvc/service/CaseProcessor.java
```diff
@@ -58,6 +53,16 @@ public Case saveCase(CreateCaseSample createCaseSample) {
     return caze;
   }
 
+  public int getUniqueCaseRef() {
```
NickGrantONS: why is this public?

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,110 @@
...
+            .ceExpectedCapacity(null)
+            .build();
+
+    caseRepository.save(individualResponseCase);
+
+    return individualResponseCase;
+  }
 
-    if (fulfilmentRequestEvent.getDateTime() == null) {
+  private void validateFulfilmentRequiredFields(
```
NickGrantONS: if we're going to start building validators we're going to be working until the end of time producing YAGNI code instead of delivering features and functionality. What is the requirement for this?

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,110 @@
...
+      EventDTO eventDTO, FulfilmentRequestDTO fulfilmentRequestDTO, String caseId) {
+    if (eventDTO.getDateTime() == null) {
       log.error(DATETIME_NOT_PRESENT);
       throw new RuntimeException(
           String.format("Date time not found in fulfilment request event for Case ID '%s", caseId));
     }
 
-    Case caze = cazeResult.get();
+    if (StringUtils.isEmpty(fulfilmentRequestDTO.getFulfilmentCode())) {
+      log.error(FULFILMENT_CODE_NOT_FOUND);
```
NickGrantONS: it's not "not found" it's "not present" i.e. null or blank, which is different. This kind of validation is nonsense... if you send us a garbage message, tough luck buddy. Otherwise we'll be spending an infinite amount of time imagining all the myriad ways that upstream could possibly mess up, which is clearly out of scope for a closed system

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,110 @@
...
 
-    eventLogger.logFulfilmentRequestedEvent(
-        caze,
-        UUID.fromString(caseId),
-        fulfilmentRequestEvent.getDateTime(),
-        FULFILMENT_REQUEST_RECEIVED,
-        FULFILMENT_REQUESTED,
-        fulfilmentRequestPayload,
-        fulfilmentRequestEvent);
+  private Case getCaseByCaseId(FulfilmentRequestDTO fulfilmentRequest) {
```
NickGrantONS: how many bazzilions of places in the code are there already similar methods as this? Also it doesn't do what it says on the method name: "find by case ID". If it's "find by case ID" then it should take "case ID" as the one and only argument

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,110 @@
...
+      caseProcessor.emitCaseCreatedEvent(individualResponseCase);
     }
+  }
+
+  private Case saveIndividualResponseCase(Case caze) {
+    Case individualResponseCase =
+        caze.toBuilder()
+            .caseId(UUID.randomUUID())
+            .caseRef(caseProcessor.getUniqueCaseRef())
+            .uacQidLinks(null)
```
NickGrantONS: why not just instantiate the object and set the things we want. Setting all these nulls is crazy

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
```diff
@@ -22,44 +28,110 @@
...
+            .events(null)
+            .createdDateTime(OffsetDateTime.now())
+            .state(CaseState.ACTIONABLE)
+            .receiptReceived(false)
+            .refusalReceived(false)
+            .addressType(HOUSEHOLD_INDIVIDUAL_RESPONSE_ADDRESS_TYPE)
```
NickGrantONS: these cases will be caught by Action Scheduler reminders etc.

src/main/java/uk/gov/ons/census/casesvc/service/CaseProcessor.java
LukeLoze1: It's reused by individual response request

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
LukeLoze1: validators removed

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
LukeLoze1: removed

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
LukeLoze1: this function is called from everywhere that does findCaseById ... now, and it's in CaseProcessor

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestProcessor.java
LukeLoze1: As discussed in standup, we believe this isn't covered by this ticket.

PR #43, opened 2019-08-20
Add Case Type to entity and default new case to household

rjweeks:
# Motivation and Context
Add new Case Type column to Case Processor and Action scheduler cases tables.
This will allow RM to differentiate a Household case from a Household individual PQ case.

# What has changed
Entity has a new CaseType field.

# How to test?
Build new image and execute in Docker Dev and run acceptance tests.

# Links
https://trello.com/c/gUMjDkBu/1064-create-case-type-variable-to-differentiate-hh-and-hi-as-opposed-to-address-type-5
ONSdigital/census-rm-ddl#6
https://github.com/ONSdigital/census-rm-action-scheduler/pull/29
https://github.com/ONSdigital/census-rm-case-api/pull/26


rjweeks:
[Create Case Type variable to differentiate HH and HI (as opposed to address type) (5)](https://trello.com/c/gUMjDkBu/1064-create-case-type-variable-to-differentiate-hh-and-hi-as-opposed-to-address-type-5)

src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
```diff
@@ -48,6 +48,7 @@ public Case saveCase(CreateCaseSample createCaseSample) {
 
     Case caze = mapperFacade.map(createCaseSample, Case.class);
     caze.setCaseRef(caseRef);
+    caze.setCaseType("HH");
```
liamtoozer: Set as `HOUSEHOLD_RESPONSE_ADDRESS_TYPE` maybe?

rjweeks: Done

src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
```diff
@@ -21,6 +21,8 @@
 
   @Column private UUID caseId;
 
+  @Column private String caseType;
```
NickGrantONS: I guess it doesn't really matter, but any reason why this is top of the list? I would've put it after addressType, or maybe at the end

src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
```diff
@@ -27,6 +27,7 @@
   private static final Logger log = LoggerFactory.getLogger(CaseService.class);
   private static final String CASE_NOT_FOUND_ERROR = "Case not found error";
   private static final String SURVEY = "CENSUS";
+  private static final String HOUSE_RESPONSE_ADDRESS_TYPE = "HH";
```
NickGrantONS: I guess it doesn't really matter, but maybe we should use "household" instead of "house" given that's the terminology used everywhere else

rjweeks: Changed.

src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
rjweeks: Moved close to addressType.

PR #44, opened 2019-08-20
Get rid of a million methods that do the same thing

NickGrantONS:
# Motivation and Context
There were 7 different versions of "log an event" when really, the class did one thing and one thing only: namely, it logged an event.

# What has changed
Got rid of the 7 different versions of "log an event". Also, generally tidied up.

# How to test?
Run the acceptance tests. There should be zero regression.

# Links
Trello: https://trello.com/c/M3dVPIVL


NickGrantONS:
[Make things better (2 hours)](https://trello.com/c/M3dVPIVL/1066-make-things-better-2-hours)

PR #52, opened 2019-08-29
Ignore RESPONDENT_AUTHENTICATED messages

NickGrantONS:
# Motivation and Context
We are receiving RESPONDENT_AUTHENTICATED on the same topic as SURVEY_LAUNCHED messages, which we can't deserialise because it's an unexpected event type, although it is in the event dictionary.

# What has changed
Ignore the RESPONDENT_AUTHENTICATED messages.

# How to test?
Publish a RESPONDENT_AUTHENTICATED event and check that Case Processor doesn't blow up.

# Links
Trello: https://trello.com/c/ZhSn40ut


NickGrantONS:
[Defect: Respondent Authenticated messages received by Survey Launched handler, because they share the same topic](https://trello.com/c/ZhSn40ut/1087-defect-respondent-authenticated-messages-received-by-survey-launched-handler-because-they-share-the-same-topic)

PR #54, opened 2019-09-02
Rmc 211 individual response mappings

LukeLoze1:
# Motivation and Context
Fix HI Case not be recorded correctly and check print files are created when requested

# What has changed
Adds individual questionnaire request print fulfilment codes to map that makes these types create and emit new cases.
Fix from previous spec to use IndividualCaseId to create the new Case

# How to test?
Uses other repos on this branch for the ticket

# Links
https://trello.com/c/KrSKaCSj/1008-rmc-211b-implement-individual-pq-requested-event-5


src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
```diff
@@ -56,15 +65,18 @@ public void processFulfilmentRequest(ResponseManagementEvent fulfilmentRequest)
...
     if (individualResponseRequestCodes.contains(fulfilmentRequestPayload.getFulfilmentCode())) {
-      Case individualResponseCase = prepareIndividualResponseCaseFromParentCase(caze);
+      Case individualResponseCase = prepareIndividualResponseCase(caze, fulfilmentRequest);
       caseService.saveAndEmitCaseCreatedEvent(individualResponseCase);
     }
   }
 
-  private Case prepareIndividualResponseCaseFromParentCase(Case parentCase) {
+  private Case prepareIndividualResponseCase(
+      Case parentCase, ResponseManagementEvent fulfilmentRequest) {
```
NickGrantONS: do you really need to pass the whole ResponseManagementEvent into this or could you just pass the individual case ID?

src/test/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestServiceTest.java
```diff
@@ -83,6 +83,8 @@ public void testGoodIndividualResponseFulfilmentRequestForUACIT4() {
     testIndividualResponseCode(HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_NORTHERN_IRELAND);
   }
 
+  //  Add mapping tests here for P_OR_X...
```
NickGrantONS: yep... you really need those tests

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
LukeLoze1: changed it without quibblizing, partly because it will need just Id when merged in with Ade's changes.  This prepare function has been moved to caseService as re-used.

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
```diff
@@ -20,17 +20,26 @@
...
-  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_WALES_WELSH = "UACIT2W";
-  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_NORTHERN_IRELAND = "UACIT4";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_ENGLAND_SMS = "UACIT1";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_WALES_ENGLISH_SMS = "UACIT2";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_WALES_WELSH_SMS = "UACIT2W";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_NI_SMS = "UACIT4";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_ENGLAND_PRINT = "P_OR_I1";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_WALES_ENGLISH_PRINT = "P_OR_I2";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_WALES_WELSH_PRINT = "P_OR_I2W";
+  private static final String HOUSEHOLD_INDIVIDUAL_RESPONSE_REQUEST_NI_PRINT = "P_OR_I4";
```
AdamHawtin: Are there enough of these now to make it worth using enums?

LukeLoze1: It's not a bad idea, but maybe not now.  Personally I wouldn't might a team discussion on these sort of things, do we bother with CONSTANT, go full hog with enums or just use the "P_OR_I..." if it's used in a named map etc.

AdamHawtin: Makes sense, I just wanted to open the debate as we're ending up with more and more of these massive boilerplate chunks of constant strings. Agree we should make a decision as a team on what practice we move forward with, this works fine for now.

PR #55, opened 2019-09-02
Undelivered as addressed

NickGrantONS:
# Motivation and Context
QM (Leidos) and PPO (Royal Mail) will receive email that is undeliverable - undelivered as addressed - and they will notify RM. We need to send a message to Field so that the cases can be followed up and we can find out why the mail couldn't be delivered.

# What has changed
Case processor marks the case as undelivered as addressed, logs a case event, and emits a case updated event.

# How to test?
Run the acceptance tests, along with Pubsub, Action Scheduler and Case API on the same branch (`undelivered-as-addressed`) along with census-rm-docker-dev.

# Links
Trello: https://trello.com/c/HOWUDNMR


ryangrundy7:
[RMC-202 Undelivered As Addressed (UAA) - FULL SOLUTION (13)](https://trello.com/c/HOWUDNMR/1059-rmc-202-undelivered-as-addressed-uaa-full-solution-13)

src/test/java/uk/gov/ons/census/casesvc/service/EventServiceTest.java
```diff
@@ -104,7 +103,7 @@ public void testProcessPrintCaseSelected() {
     // Given
     EasyRandom easyRandom = new EasyRandom();
     Case caze = easyRandom.nextObject(Case.class);
-    when(caseService.findCase(anyInt())).thenReturn(Optional.of(caze));
+    when(caseService.getCaseByCaseRef(anyInt())).thenReturn(caze);
```
LukeLoze1: caseRef can be set here rather than 123,  otherwise the getting caze from caseRef isn't tested in the unit test

NickGrantONS: there is adequate test coverage. Have amended anyway but probably not a useful change at all.

PR #57, opened 2019-09-05
Log FIELD_CASE_SELECTED events

NickGrantONS:
# Motivation and Context
When Action Scheduler sends cases to Fieldwork for followup, we need to notify the Case Processor so that it can log an event against the case.

# What has changed
Receive a messages from Action when cases are sent to field, and log the event.

# How to test?
Run the acceptance tests, along with Case API and Action Scheduler on the branch `log-action-field-events`.

# Links
Trello: https://trello.com/c/VaYONhzW


NickGrantONS:
[RMC-133 (part 2) Update case history when cases scheduled for visits are sent to Field (8)](https://trello.com/c/VaYONhzW/927-rmc-133-part-2-update-case-history-when-cases-scheduled-for-visits-are-sent-to-field-8)

PR #59, opened 2019-09-06
Set unreceipt flag to Boolean to stop being stored in event payload json

hbrace:
# Motivation and Context

Currently, the json payload in a Survey Launched event contains unreceipt: false. This should not be there.

# What has changed

The unreceipt type has been changed to Boolean (from boolean) to allow nulls so that it can be excluded from the json.

# How to test?

Run acceptance tests. Have a look in your DB for a survey launched event and see if "unreceipt": false is not present.

# Links

https://trello.com/c/H2V1Mb8f/1111-survey-launched-event-storing-unreceipt-flag-in-json-payload-in-case-events

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/ResponseDTO.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/ResponseDTO.java
@@ -16,7 +16,8 @@ public class ResponseDTO {
 
   private String questionnaireId;
 
-  private boolean unreceipt;
+  @JsonInclude(Include.NON_NULL)
+  private Boolean unreceipt;
 
   private String agentId;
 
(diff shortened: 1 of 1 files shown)


hbrace:
[Survey launched event storing Unreceipt flag in JSON payload in case events](https://trello.com/c/H2V1Mb8f/1111-survey-launched-event-storing-unreceipt-flag-in-json-payload-in-case-events)

src/main/java/uk/gov/ons/census/casesvc/model/dto/ResponseDTO.java
```diff
@@ -16,7 +16,8 @@
 
   private String questionnaireId;
 
-  private boolean unreceipt;
+  @JsonInclude(Include.NON_NULL)
```
AdeHarrison: could be put at class level requiring single annotation

AdeHarrison:
not a blocker but you could add validation for the event_payload field, RE: QuestionnaireLinkedReceiverIT - the contents have caused quite a few problems

src/main/java/uk/gov/ons/census/casesvc/model/dto/ResponseDTO.java
hbrace: Done 👀

PR #60, opened 2019-09-09
use caseType not addressType to set the caseType field in uac_updated message

davidmort:
# Motivation and Context
We are currently using the addressType field from case data to populate the caseType field that is part of the uac_updated event that gets emitted, this should be using the recently added caseType field

# What has changed
Changed code to set the caseType field on the uac_updated event from the caseType field from the case data

# How to test?
- run acceptance tests


davidmort:
[UAC updated event for HI cases emitting caseType as HH](https://trello.com/c/UIP64Z8X/1133-uac-updated-event-for-hi-cases-emitting-casetype-as-hh)

PR #62, opened 2019-09-11
Log poison messages

NickGrantONS:
# Motivation and Context
We have no idea why Spring was unable to process JSON messages which are malformed or unmarshallable, and we have no log output to see which the offending message is, which makes operational support really difficult.

# What has changed
Added a global error handler for bad messages and checked JSON validity before logging meaningful error output and the offending message payload.

# How to test?
Publish a dodgy message. Check the logs.

# Links
Trello: https://trello.com/c/eXQ4PkKa


NickGrantONS:
[Tech must have: improved error logging for bad messages](https://trello.com/c/eXQ4PkKa/1150-tech-must-have-improved-error-logging-for-bad-messages)

PR #66, opened 2019-09-19
change config to make all rabbit channels transactional

davidmort:
# Motivation and Context
All rabbit channels should be transactional

# What has changed
updated the rabbit config in AppConfig.java to make sure all rabbit channels are transactional

# How to test?
start up service, go to the "channels" tab in the management UI and all channels from this app should be marked as transactional ("T" in the mode column)


davidmort:
[make all rabbit channels transactional (3)](https://trello.com/c/2tn45jHY/1180-make-all-rabbit-channels-transactional-3)

PR #71, opened 2019-10-07
1099 css with qid

LukeLoze1:
# Motivation and Context
qid linking with CCSCaseListed

# What has changed
doesn't send out event if qid provided, tests for this

# How to test?
Install locally and try with AT branch: https://github.com/ONSdigital/census-rm-acceptance-tests/pull/137

https://trello.com/c/KNsTMSMf/1099-rmc-1663-handle-qid-linking-in-ccspropertylisted-event-5

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
@@ -9,26 +9,29 @@ import uk.gov.ons.census.casesvc.model.dto.ResponseManagementEvent;
 import uk.gov.ons.census.casesvc.model.entity.Case;
 import uk.gov.ons.census.casesvc.model.entity.EventType;
 import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
+import uk.gov.ons.census.casesvc.model.repository.UacQidLinkRepository;
 
 @Service
 public class CCSPropertyListedService {
-
   private static final String CCS_ADDRESS_LISTED = "CCS Address Listed";
 
   private final UacService uacService;
   private final EventLogger eventLogger;
   private final CaseService caseService;
   private final CcsToFieldService ccsToFieldService;
+  private final UacQidLinkRepository uacQidLinkRepository;
 
   public CCSPropertyListedService(
       UacService uacService,
       EventLogger eventLogger,
       CaseService caseService,
-      CcsToFieldService ccsToFieldService) {
+      CcsToFieldService ccsToFieldService,
+      UacQidLinkRepository uacQidLinkRepository) {
     this.uacService = uacService;
     this.eventLogger = eventLogger;
     this.caseService = caseService;
     this.ccsToFieldService = ccsToFieldService;
+    this.uacQidLinkRepository = uacQidLinkRepository;
   }
 
   public void processCCSPropertyListed(ResponseManagementEvent ccsPropertyListedEvent) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
```diff
@@ -37,7 +38,18 @@ public void processCCSPropertyListed(ResponseManagementEvent ccsPropertyListedEv
...
+      uacService.createUacQidLinkedToCCSCase(caze);
+    } else {
+      UacQidLink uacQidLink = uacService.findByQid(ccsProperty.getUac().getQuestionnaireId());
+      uacQidLink.setCaze(caze);
+
+      if (uacQidLink == null) {
+        throw new RuntimeException(
+            "Received unknown QID for CSSProperty Listed Event: "
+                + ccsProperty.getUac().getQuestionnaireId());
+      }
```
AdeHarrison: move above "uacQidLink.setCaze(caze);" as could be null

AdeHarrison: need unit test coverage for this condition also

src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
```diff
@@ -47,6 +59,8 @@ public void processCCSPropertyListed(ResponseManagementEvent ccsPropertyListedEv
         ccsPropertyListedEvent.getEvent(),
         convertObjectToJson(ccsProperty));
 
-    ccsToFieldService.convertAndSendCCSToField(caze);
+    if (ccsProperty.getUac() == null) {
+      ccsToFieldService.convertAndSendCCSToField(caze);
```
AdeHarrison: no need to check for null if moved into "if (ccsProperty.getUac() == null) {" condition above

src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
```diff
@@ -37,7 +41,14 @@ public void processCCSPropertyListed(ResponseManagementEvent ccsPropertyListedEv
         caseService.createCCSCase(
             ccsProperty.getCollectionCase().getId(), ccsProperty.getSampleUnit());
 
-    uacService.createUacQidLinkedToCCSCase(caze);
+    if (ccsProperty.getUac() == null) {
+      uacService.createUacQidLinkedToCCSCase(caze);
+      ccsToFieldService.convertAndSendCCSToField(caze);
+    } else {
+      UacQidLink uacQidLink = uacService.findByQid(ccsProperty.getUac().getQuestionnaireId());
```
NickGrantONS: I can absolutely guarantee that the requirements will change and it'll be a case of "oh we sometimes don't bother linking and maybe we'll do it later" in which case this would error I guess... but I support what you've done here. It meets the current requirements as we understand them

src/test/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedServiceTest.java
```diff
@@ -83,4 +89,44 @@ public void testGoodCCSPropertyListed() {
...
+    when(caseService.createCCSCase(
+            expectedCaseId, responseManagementEvent.getPayload().getCcsProperty().getSampleUnit()))
+        .thenReturn(expectedCase);
+
+    underTest.processCCSPropertyListed(responseManagementEvent);
+
+    ArgumentCaptor<Case> caseCaptor = ArgumentCaptor.forClass(Case.class);
+    verify(eventLogger)
+        .logCaseEvent(
+            caseCaptor.capture(),
```
AdeHarrison: captured but not tested

PR #75, opened 2019-10-10
Use Exception Manager to reduce logging, allow message peeking and skipping

NickGrantONS:
# Motivation and Context
To be able to skip messages on Rabbit without any downtime, we need the Exception Manager.

# What has changed
Added Exception Manager client and created a new `MessageErrorHandler` which handles reduced logging, message 'peeking', message skipping etc. etc.

# How to test?
Run with the Exception Manager and then publish some bad messages.

# Links
[Trello](https://trello.com/c/f7LmzvJm)


NickGrantONS:
[SPIKE: zero downtime message quarantine and error log reduction (5 days)](https://trello.com/c/f7LmzvJm/1244-spike-zero-downtime-message-quarantine-and-error-log-reduction-5-days)

rjweeks:
Should we use a circuit breaker? https://microservices.io/patterns/reliability/circuit-breaker.html

NickGrantONS:
> Should we use a circuit breaker? https://microservices.io/patterns/reliability/circuit-breaker.html

No, circuit breakers are to stop systems which are mostly synchronous RESTful API calls, ending up DDoSing themselves. The only API calls we make are to Case API and Gov Notify, and neither of those are done from the Case Processor. We could use the Exception Manager to rate limit a little bit, but the goals are pretty straightforward for the Exception Manager:
 1. Reduce log spew
 2. Gather all the message hashes of the bad messages without having to scrape the logs
 3. Allow us to 'peek' at the message contents without having to stop the queue consumer
 4. Allow us to skip a message without having to stop the queue consumer
 6. Allow all of the above items to be easily automated by some kind of tool/script/wizard/UI, so that the job of operationally sorting out these messes is a lot easier

Because we don't have the A -> B -> C -> D chain of REST API calls that are common in other systems, given that we're message-driven, as opposed to responding to synchronous API requests, means that tracing a 'call stack' across a distributed system is less of a problem, and the only retries are on messages, which spams our DB a little bit (e.g. looking for a nonexistent case or QID) but it's naturally limited by the number of queue consumers. The circuit breakers are usually there because people start clicking 'refresh' loads when a website is struggling, which exacerbates problems. RM is pretty much beautifully decoupled - we can stop almost any part of the system and it doesn't cause much of a problem except log spew, which the Exception Manager takes care of.

It would be ideal to have transaction IDs consistently used across RM... but that's covered by a different ticket.

I know it feels a little wrong to be hand-rolling a solution, but trust me: we don't want to go down the DLQ road, because it's a nightmare to sort out a major incident - so many messages to replay from so many queues, and no way of knowing which messages are bad, and which are not.

All I can say is... the Exception Manager is a pretty simple beast, so we should try it. If it improves our lives and it's robust and reliable, then it was the right thing to do. Back to the drawing board if it doesn't do what it says on the tin.

rjweeks:
ok, thanks. I do trust you.

rjweeks:
I'm ok with this @davidmort

NickGrantONS:
Starting a new branch / PR with just these changes... to avoid releasing CCS stuffs

PR #76, opened 2019-10-11
reject link event where uacqid already linked to different case

LukeLoze1:
# Motivation and Context
https://trello.com/c/4QGNALnV/1258-defect-can-re-link-a-uac-qid-and-a-case-3

# What has changed
If a qidLink comes in but that qidlink is already linked to another case it shouldn't relink it to the different case but also throw a meaningful Exception

# How to test?
Run locally, there's no ATs or ITs just new Unit tests.  Everything else, ITs & ATs should continue to work


src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
```diff
@@ -59,4 +61,12 @@ public void processQuestionnaireLinked(ResponseManagementEvent questionnaireLink
         questionnaireLinkedEvent.getEvent(),
         convertObjectToJson(uac));
   }
+
+  private void checkQidNotLinkedToAnotherCase(UacDTO uac, UacQidLink uacQidLink) {
+    if (uacQidLink.getCaze() != null) {
+      if (!uacQidLink.getCaze().getCaseId().equals(UUID.fromString(uac.getCaseId()))) {
```
NickGrantONS: what's wrong with an `&&` and combining these two conditions into a single if statement?

src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
```diff
@@ -59,4 +61,12 @@ public void processQuestionnaireLinked(ResponseManagementEvent questionnaireLink
...
+        throw new RuntimeException("UacQidLink already linked to a different case");
```
NickGrantONS: Already linked to _which_ case? Put the case IDs in the exception so that we have something to go on.

PR #81, opened 2019-10-16
Always create uac-qid pair for CCS property listed event.

rjweeks:
# Motivation and Context
Always create a new uac-qid pair when a CCS property listed event is received. This is because a CCS PQ may be linked but not completed and if a field agent calls on property again they will need a uac-qid to launch an EQ.

# What has changed
Small code change to always create uac-qid

# How to test?
Build locally and run in docker-dev then run acceptance tests

# Links
https://trello.com/c/TRtVJEzn/1268-qid-linked-in-property-listing-event-should-create-a-uac-qid-pair-by-default-on-case-creation-3

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CCSPropertyListedService.java
@@ -47,11 +47,13 @@ public class CCSPropertyListedService {
             isRefused,
             isInvalidAddress);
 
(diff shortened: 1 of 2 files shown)


rjweeks:
[QID linked in Property Listing event should create a UAC/QID pair by default on case creation (3)](https://trello.com/c/TRtVJEzn/1268-qid-linked-in-property-listing-event-should-create-a-uac-qid-pair-by-default-on-case-creation-3)

PR #85, opened 2019-10-30
Hotfix to v 2 8 6 fix msg persistence

LukeLoze1:
# Motivation and Context
Msgs sent to delay or delivery weren't set as Persistent, so on a rabbit restart they'd be lost

# What has changed
Make msgs sent to delay or quarantine Persistent, deliveryMode wasn't set by default.
Also removed unneeded bool settings around result.

The only changes to the base branch v.2.8.6 is adding
     // At this point message is not persistent, we need it to be persistent
      message.getMessageProperties().setDeliveryMode(MessageDeliveryMode.PERSISTENT); 
twice

# How to test?
Build this branch locally, tag it & push to your GCP.  Using rabbit GUI send 2 bad msgs to case-processor, i.e. case.ccsPropertyListedQueue.  Then with the toolbox use msgWizard to quarantine 1 of the msgs.  If you now look on the delayedRedeliveryQueue & the quarantineQueue there should be 1 msg on each marked persistent.

Then in your workloads drill into rabbitmq and scale down to 0 from 3.  Then when completed scaling down, scale back up to 3.  Wait till all green then navigate back to the RabbitUI,  there should be a persistent msg in Quarantine and 1 on the delayedRedeliveryQueue queue, unless it's already been sent back to case.ccsPropertyListedQueue, which is fine


src/main/java/uk/gov/ons/census/casesvc/messaging/ManagedMessageRecoverer.java
```diff
@@ -0,0 +1,229 @@
...
+              exceptionManagerClientException);
+    }
+
+    // Check if OK and the message is stored... then we can go ahead and quarantine
+    if (result) {
+      result = false; // The next bit might go wrong
+      log.with("message_hash", messageHash).warn("Skipping message");
+
+      // At this point message is not persistent, we need it to be persistent
+      message.getMessageProperties().setDeliveryMode(MessageDeliveryMode.PERSISTENT);
```
NickGrantONS: This is one of the two actual tiny changes

src/main/java/uk/gov/ons/census/casesvc/messaging/ManagedMessageRecoverer.java
```diff
@@ -0,0 +1,229 @@
...
+          reportResult, messageHash, rawMessageBody, listenerExecutionFailedException, message)) {
+        return; // Our work here is done
+      }
+
+      peekMessage(reportResult, messageHash, rawMessageBody);
+      logMessage(
+          reportResult, listenerExecutionFailedException.getCause(), messageHash, rawMessageBody);
+
+      // At this point message is not persistent, we need it to be persistent
+      message.getMessageProperties().setDeliveryMode(MessageDeliveryMode.PERSISTENT);
```
NickGrantONS: This is one of the two actual tiny changes

LukeLoze1:
> 👍 But if we merge #85 first and rebase this once it's been released, this should boil down to no changes?

PR #86, opened 2019-11-01
Making dockerfile non root

ryangrundy7:
# Motivation and Context
Running docker containers in root is a security concern so there should be a user that's created that the container can use instead. 

# What has changed
- Added a non-root user
- Updated dockerfile layout
# How to test?
- Run mvn clean install
- Run in docker dev environment with other prs
- Run in kubernetes environment with other prs

# Links
[Trello](https://trello.com/c/dUT7y0TW)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,12 @@
 FROM openjdk:11-jdk-slim
 
 ARG JAR_FILE=census-rm-case-processor*.jar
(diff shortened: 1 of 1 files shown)


ryangrundy7:
[apps running as non-root (8)](https://trello.com/c/dUT7y0TW/525-apps-running-as-non-root-8)

PR #92, opened 2020-01-08
469 using rabbit timestamp

ryangrundy7:
# Motivation and Context
The aim of the PR is to store the timestamp that comes in on the rabbit message headers. We're doing this because the datetime field we store can sometimes be potentially 24 hours out because it's relying on external systems/processes.
# What has changed
- Added `messageTimestamp` in the event model
- updated all receivers and services to get the rabbit timestamp from its headers.
- Updated all tests.

# How to test?
- Run `mvn clean install` and run all the tests.
- Run with the acceptance tests.

# Links
[Trello](https://trello.com/c/0mdJC6nl/)


ryangrundy7:
[Capture Rabbit Msg Timestamp in case-processor (5)](https://trello.com/c/0mdJC6nl/469-capture-rabbit-msg-timestamp-in-case-processor-5)

src/main/java/uk/gov/ons/census/casesvc/utility/MsgDateHelper.java
```diff
@@ -0,0 +1,19 @@
...
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import org.springframework.messaging.Message;
+
+public class MsgDateHelper {
+  public static OffsetDateTime getMsgTimeStamp(Message<?> msg) {
+
+    if (msg.getHeaders().getTimestamp() == null) {
+      throw new RuntimeException(
+          "Message Headers missing Timestamp, is this important enough to reject message on?");
```
AdamHawtin: Looks like a work in progress exception message, we need to decide if it is important enough to reject the message.

PR #96, opened 2020-01-17
fix sample loading for ceExpectedCapacity

hbrace:
# Motivation and Context

When a sample unit has a blank CE_EXPECTED_CAPACITY in the sample file, case processor cannot ingest it and rejects the message. This change is to enable case processor to accept these sample units. 

# What has changed

ceExpectedCapacity changed to Integer in CreateCaseSample.
Unit test updated to allow this and is also checked in an integration test.

# How to test?

Maven clean install
Run [acceptance tests on branch of the same name](https://github.com/ONSdigital/census-rm-acceptance-tests/pull/157) - they should pass.
Run performance tests - they should pass now.
Push case processor image to your GCP if you fancy, then run acceptance tests and performance tests.

# Links

https://trello.com/c/iPx9qCU5/495-fix-sample-files-or-sample-file-loading-for-ceexpectedcapacity-3

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/CreateCaseSample.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/CreateCaseSample.java
@@ -29,7 +29,7 @@ public class CreateCaseSample {
   private String fieldCoordinatorId;
   private String fieldOfficerId;
   private String treatmentCode;
-  private String ceExpectedCapacity;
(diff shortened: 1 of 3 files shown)


hbrace:
[Fix sample files or sample file loading for ceExpectedCapacity (3)](https://trello.com/c/iPx9qCU5/495-fix-sample-files-or-sample-file-loading-for-ceexpectedcapacity-3)

PR #98, opened 2020-01-20
Add last updated column to case entity

hbrace:
# Motivation and Context

Adding this lastUpdated time stamp will provide the capability to do 'delta' based extracts from RM for data that is sent to DAP for use by processing and the RCA/FPA algorithms as opposed to exported the full set each time.

# What has changed

lastUpdated column added to Case entity, this is annotated so that the lastUpdated time is updated when the row is updated.
An integration test has been modified to check that the lastUpdated is different from when a case is created and it is updated with a refusal event.

# How to test?

Build case processor, run acceptance tests. 
Get some data in your database and check the last_updated column for a specific case. Submit a message to rabbitmq that will update this case (refusal, receipt etc) and check the last_updated field has changed.

# Links

https://trello.com/c/YHny5NEN/223-add-lastupdated-field-to-case-db-cases-table-3

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
@@ -13,6 +13,7 @@ import javax.persistence.Table;
 import lombok.Data;
 import org.hibernate.annotations.Generated;
 import org.hibernate.annotations.GenerationTime;
+import org.hibernate.annotations.UpdateTimestamp;
 
 @Data
 @Entity
@@ -115,4 +116,6 @@ public class Case {
 
   @Column(nullable = false, columnDefinition = "BOOLEAN DEFAULT false")
   private boolean undeliveredAsAddressed;
+
+  @Column @UpdateTimestamp private OffsetDateTime lastUpdated;
 }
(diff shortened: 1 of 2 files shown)


hbrace:
[add "lastUpdated" field to Case DB cases table (3)](https://trello.com/c/YHny5NEN/223-add-lastupdated-field-to-case-db-cases-table-3)

NickGrantONS:
if we're going to use Hibernate features like this, maybe we should use `@CreationTimestamp` too?

src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
```diff
@@ -115,4 +116,6 @@
 
   @Column(nullable = false, columnDefinition = "BOOLEAN DEFAULT false")
   private boolean undeliveredAsAddressed;
+
+  @Column @UpdateTimestamp private OffsetDateTime lastUpdated;
```
NickGrantONS: minor comment, but because this isn't a `timestamp with time zone` the value returned by the database cannot be guaranteed to be UTC. Strictly speaking, somebody querying our DB from a different time zone might see the incorrect time - it might cause us problems with BST.

hbrace: Added in timestamp with timezone column definition

PR #100, opened 2020-01-29
Refactor to handle new UAC-QIDs as part of fulfilment messages

NickGrantONS:
# Motivation and Context
The order that we received the UAC created and fulfilment messages resulted in either success or failure. Given that we have no control over which message we process first, we needed to refactor.

# What has changed
Combined the UAC-QID created details into the fulfilment message, so it can be linked to a new case, should one need to be created.

# How to test?
Run the acceptance tests. There should be zero regression (and a whole lot less errors in the logs).

# Links
Trello: https://trello.com/c/yNjry5aD


NickGrantONS:
[TECH DEBT: Telephone Capture Fulfilment Refactoring (8)](https://trello.com/c/yNjry5aD/530-tech-debt-telephone-capture-fulfilment-refactoring-8)

PR #104, opened 2020-02-06
add new caseReceipter service to be shared between receiptprocessor a…

LukeLoze1:
…nd questionnaire linker

# Motivation and Context
Questionnaire Linking wasn't working with continuation receipting (was receipting them) and will need to use same logic as receiptReceiver when deciding on Receipting, actualResponses incrementing and so on

# What has changed
New Service called CaseReceipter,  takes the Case and Uac and decides whether or not to receipt them.  Very limited new functionality, apart from Questionnaires linked on continuation qids not causing a receipt anymore. 

# How to test?
Should work with mvn clean install, and with everything else on Master with the ATs

# Links
https://trello.com/c/4Zv7UzFp/571-ce-and-spg-receipting-refactor-linking-to-use-new-receipt-logic-8

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
@@ -1,0 +1,27 @@
+package uk.gov.ons.census.casesvc.service;
+
+import static uk.gov.ons.census.casesvc.utility.QuestionnaireTypeHelper.iscontinuationQuestionnaireTypes;
+
+import org.springframework.stereotype.Component;
+import uk.gov.ons.census.casesvc.model.entity.Case;
+import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
+
+@Component
+public class CaseReceipter {
+  private final CaseService caseService;
+
+  public CaseReceipter(CaseService caseService) {
+    this.caseService = caseService;
+  }
+
+  public void handleReceipting(Case caze, UacQidLink uacQidLink) {
+    if (uacQidLink.isActive()) return;
+
+    if (caze.isReceiptReceived()) return;
+
+    if (!iscontinuationQuestionnaireTypes(uacQidLink.getQid())) {
+      caze.setReceiptReceived(true);
+      caseService.saveAndEmitCaseUpdatedEvent(caze);
+    }
+  }
+}
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
@@ -15,8 +15,6 @@ public class CaseReceipter {
   }
 
   public void handleReceipting(Case caze, UacQidLink uacQidLink) {
-    if (uacQidLink.isActive()) return;
-
     if (caze.isReceiptReceived()) return;
 
     if (!iscontinuationQuestionnaireTypes(uacQidLink.getQid())) {
--- a/src/main/java/uk/gov/ons/census/casesvc/messaging/ReceiptReceiver.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/messaging/ReceiptReceiver.java
@@ -8,20 +8,20 @@ import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.Message;
 import org.springframework.transaction.annotation.Transactional;
 import uk.gov.ons.census.casesvc.model.dto.ResponseManagementEvent;
-import uk.gov.ons.census.casesvc.service.ReceiptService;
+import uk.gov.ons.census.casesvc.service.QidReceiptService;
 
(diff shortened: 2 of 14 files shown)


src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.ons.census.casesvc.service;
+
+import static uk.gov.ons.census.casesvc.utility.QuestionnaireTypeHelper.iscontinuationQuestionnaireTypes;
+
+import org.springframework.stereotype.Component;
+import uk.gov.ons.census.casesvc.model.entity.Case;
+import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
+
+@Component
+public class CaseReceipter {
```
AdamHawtin: Is there a reason this can't live in the ReceiptService?

src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
```diff
@@ -0,0 +1,27 @@
...
+  private final CaseService caseService;
+
+  public CaseReceipter(CaseService caseService) {
+    this.caseService = caseService;
+  }
+
+  public void handleReceipting(Case caze, UacQidLink uacQidLink) {
+    if (uacQidLink.isActive()) return;
```
AdamHawtin: I presume this check is because the flow we want is to always deactivate the UAC before calling this function, but if we're doing that think it should actively throw an exception instead of just returning if that's not the followed, maybe throw an `IllegalArgumentException`?

src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
LukeLoze1: It could, I was in favour of sticking it in the ReceiptService to start with.  However when I started on it , it felt nicer to stick it in it's own clean class.  This keeps a cleanish pattern, where both the ReceiptService & QuestionnaireLinkedService follow the same pattern lightweight pattern.  

The 'CaseReceipter' will be easier to test for both standard receipts and linking events, otherwiseI you'd want to test it through the public 'processReceipt' which would call the 'handleReceipting' function,  which is a bit more effort setting up more mocks and messages for each combination.  Then would you have to retest the handleReceipting for direct calls from the Linker?

src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
LukeLoze1: Good spot, I think this should be moved back to the Linking Service where it might not be inactive.  Making it inactive is what the receiptService does, so silly to test for it further along

src/main/java/uk/gov/ons/census/casesvc/service/CaseReceipter.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Component
+public class CaseReceipter {
+  private final CaseService caseService;
+
+  public CaseReceipter(CaseService caseService) {
+    this.caseService = caseService;
+  }
+
+  public void handleReceipting(Case caze, UacQidLink uacQidLink) {
```
NickGrantONS: is it reasonable to presume that the uacQidLink is always synonymous with the case being receipted? Why isn't the `active` flag being set to false on the `UacQidLink`?

AdamHawtin: The idea was this method could be used for both new incoming receipts and receipting a case after a link comes in to an already receipted UacQidLink, in the second case we don't need to update the UacQidLink again (although is there any harm in setting the `active` flag false again?)

If this method is only going to handle the case specific stuff, I think it's name should reflect that at least.

NickGrantONS: there are 5 lines of code in this function. Is this really achieving much?

AdamHawtin: I think so, logic for receipting a case is getting more complex with incoming cards, doing this now puts that in one place where it'll be applied properly on new receipts and linking to already receipted QID's

LukeLoze1: As Adam says we're purposefully moving all the logic to one place because we know that it's going to get more complex.  This will make it much easier to reuse & test.  

Naming wise I'll rename the function to 'ReceiptCase' ?

LukeLoze1:
> I question the value in having a whole extra Spring component, replete with all the wiring gubbins, just for one 5-line function, which isn't really self descriptive. I think proliferating lots of tiny Spring components isn't going to improve the codebase

Doing it in a separate component isn't very costly, and it'll make the testing a lot cleaner when it gets more complex, which we know it's going to

PR #108, opened 2020-02-18
Log establishment non-field refusals

AdamHawtin:
# Motivation and Context
We want to record establishment address level case refusals that come from channels other than field but not update the case.

# What has changed
* Log estab non field refusals
* Update test

# How to test?
Build and run acceptance tests

# Links
https://trello.com/c/FcbMllGI/633-log-estab-individual-refusal-on-ce-and-spg-3


AdamHawtin:
[Log Estab individual refusal on CE and SPG (3)](https://trello.com/c/FcbMllGI/633-log-estab-individual-refusal-on-ce-and-spg-3)

PR #109, opened 2020-02-18
Allow null region in address DTO builder

AdamHawtin:
# Motivation and Context
We were getting null pointer exceptions because the address builder relied on region being set

# What has changed
* Allow null region in address DTO builder
* Add unit test

# How to test?
Build and run acceptance tests

# Links
https://trello.com/c/1F5lJNLo/630-ccs-refusal-messages-fail-in-case-processor


AdamHawtin:
[CCS Refusal messages fail in case processor](https://trello.com/c/1F5lJNLo/630-ccs-refusal-messages-fail-in-case-processor)

PR #111, opened 2020-02-21
Added derived form type for the amusement of RH

NickGrantONS:
# Motivation and Context
RH wanted a value which could be derived from another value, so we implemented this in RM instead of RH.

# What has changed
Derived the form type from the questionnaire type. Sent it in the UACUpdated message to RH.

# How to test?
Run the acceptance tests.

# Links
Trello: https://trello.com/c/nGDQ3eJF

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/UacService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/UacService.java
@@ -103,6 +103,10 @@ public class UacService {
     uac.setUacHash(Sha256Helper.hash(uacQidLink.getUac()));
     uac.setUac(uacQidLink.getUac());
     uac.setActive(uacQidLink.isActive());
+
+    // It's perfectly possible to derive the Form Type from the supplied data, so RM should not be
+    // forced to incorporate CENSUS business logic. It's for the CENSUS team to put business logic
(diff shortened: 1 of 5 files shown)


NickGrantONS:
[UACUpdated event - Form type (8)](https://trello.com/c/nGDQ3eJF/496-uacupdated-event-form-type-8)

src/main/java/uk/gov/ons/census/casesvc/service/UacService.java
```diff
@@ -152,4 +156,28 @@ public UacQidLink findByQid(String questionnaireId) {
 
     return uacQidLinkOpt.get();
   }
+
+  private String mapQuestionnaireTypeToFormType(String qid) {
```
LukeLoze1: Not sure what our policy is on testing mappings like this? The current unit tests only check that questionnaireType 1  maps correctly

NickGrantONS: I thought about doing an exhaustive unit test and then I thought "I can't be bothered to test 15 lines of switch statement which is just about the simplest possible code I've ever written", especially as a private method. We can have exhaustive AT scenarios if people are going to go mad with testing every combination and permutation imaginable, but does the unit test give us some degree of regression protection? Yes.

PR #115, opened 2020-02-25
Remove direct link between Case Processor and Fieldwork Adapter

NickGrantONS:
# Motivation and Context
We are moving towards the Fieldwork Adapter processing the normal RM events, instead of having its own direct message which instructs it what to do. In time, this might give FWMT the ability to do FieldworkAdapter's work, and we can remove FieldworkAdapter.

# What has changed
Removed any and every reference to the direct message that used to go from Case Processor to Fieldwork Adapter.

Publish a CASE_CREATED event instead.

# How to test?
Run the acceptance tests. There should be zero regression

# Links
Trello: https://trello.com/c/ONAETtD2


NickGrantONS:
[Refactor CCSPropertyListed features (5)](https://trello.com/c/ONAETtD2/512-refactor-ccspropertylisted-features-5)

src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
```diff
@@ -105,24 +105,35 @@ public PayloadDTO saveCaseAndEmitCaseCreatedEvent(Case caze) {
     return saveCaseAndEmitCaseCreatedEvent(caze, null);
   }
 
+  public PayloadDTO saveCaseAndEmitCaseCreatedEvent(Case caze, Metadata metadata) {
```
LukeLoze1: Not asking for this to be changed for this ticket, as it's not the fault of this ticket - but I've got reservations about the size and scope of the CaseService class.  It seems to be doing quiet a lot of sort of related stuff that might, might benefit from refactoring to make it more maintainable

NickGrantONS: I agree that there seems to have been a massive proliferation of overloaded methods, and I don't know why. It makes it quite hard to trace what the heck is going on. On the plus side though, we do generally only have to change stuff in one place, I found while doing this PR

PR #119, opened 2020-03-03
LT suggestions that I overlooked on last PR

LeoKHoward:
# Motivation and Context
Merged last case-processor PR without amending it as per Liam's suggestions

# What has changed
Specified UUID.class in QuestionnaireLinkedServiceTest
Assigned name for UUID.randomUUID() in QuestionnaireLinkedService

# How to test?
No changes to any functionality but clone branch and run ```mvn clean install``` to check it builds

https://trello.com/c/pKUaVF0i

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
@@ -15,7 +15,7 @@ import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
 
 @Service
 public class QuestionnaireLinkedService {
-
+  private final UUID TEST_CASE_ID_1 = UUID.randomUUID();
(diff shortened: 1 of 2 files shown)


LeoKHoward:
[Refactored changes in case-processor](https://trello.com/c/pKUaVF0i/670-refactored-changes-in-case-processor)

AdamHawtin:
Shouldn't break anything now, however I'm not sure we need the extra line to generate the UUID? I'm personally fine with doing these things inline as we're all using IDE's that should make it easy enough to work out what that parameter is doing.

liamtoozer:
> Shouldn't break anything now, however I'm not sure we need the extra line to generate the UUID? I'm personally fine with doing these things inline as we're all using IDE's that should make it easy enough to work out what that parameter is doing.

Yeah makes sense. I don't feel strongly about it enough to leave it in but I guess it's helpful to see what it's doing at a glance (as I'm lazy!). Leave it as it was 👍

PR #121, opened 2020-03-05
Add missing indexes and fix other DDL mismatch

NickGrantONS:
# Motivation and Context
Hibernate does not set up the same database config in a dev GCP env as it does in an env built from the DDL scripts.

# What has changed
Fixed mismatches.

# How to test?
Build and run into empty database. Using a **reliable** DB inspection tool (i.e. not DataGrip) check the missing indexes are present.

# Links
Trello: https://trello.com/c/ybEYuI66

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/entity/Case.java
@@ -6,6 +6,7 @@ import java.util.UUID;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.Id;
+import javax.persistence.Index;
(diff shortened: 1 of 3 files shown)


NickGrantONS:
[DDL and Hibernate mismatch (5)](https://trello.com/c/ybEYuI66/657-ddl-and-hibernate-mismatch-5)

NickGrantONS:
> Something to consider, we're discussing removing the indexes for certain processes to speed up writes, moving them into code here will stop the case processor starting up if those indexes don't exist without a code change or turning off verify.

I disagree... I don't think Hibernate will verify the presence of the indexes, so we should be fine to start up.

PR #123, opened 2020-03-10
Add CE Secure Marker

LeoKHoward:
# Motivation and Context
New column for identifying if CE case is a Secure Establishment

# What has changed
New Metadata column which will be used to group CE/SPG details that aren't constants in the JSON
Tests

# How to test?
Clone the branch and run ```maven clean install``` which will build the image

# Links
https://collaborate2.ons.gov.uk/confluence/display/CRM/Address+Frame+specification+2021?focusedCommentId=36183836#comment-36183836
https://trello.com/c/BAI9qOOC/619-rmc-341-add-secure-marker-case-api-and-field-8

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
@@ -17,6 +17,7 @@ import uk.gov.ons.census.casesvc.model.dto.PayloadDTO;
 import uk.gov.ons.census.casesvc.model.dto.ResponseManagementEvent;
 import uk.gov.ons.census.casesvc.model.dto.SampleUnitDTO;
 import uk.gov.ons.census.casesvc.model.entity.Case;
+import uk.gov.ons.census.casesvc.model.entity.CaseMetadata;
 import uk.gov.ons.census.casesvc.model.repository.CaseRepository;
 import uk.gov.ons.census.casesvc.utility.CaseRefGenerator;
 import uk.gov.ons.census.casesvc.utility.EventHelper;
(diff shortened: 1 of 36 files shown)


LeoKHoward:
[RMC-341 Add secure marker - case api and field (8)](https://trello.com/c/BAI9qOOC/619-rmc-341-add-secure-marker-case-api-and-field-8)

src/main/java/uk/gov/ons/census/casesvc/service/CaseService.java
```diff
@@ -79,6 +77,13 @@ public Case saveCaseSample(CreateCaseSample createCaseSample) {
     caze.setCeActualResponses(0);
     caze.setHandDelivery(isTreatmentCodeDirectDelivered(createCaseSample.getTreatmentCode()));
 
+    if (caze.getCaseType().equals("CE")) {
```
AdamHawtin: Not particularly keen on building this logic specifically for CE's, I thought our general policy was to avoid conditioning on things like case type unless the features really require it? (to clarify though, I do definitely like the idea of storing sparse columns like this into a metadata json column, I just don't think we should build it to be specific to any particular case type).

LukeLoze1: I agree.  I thought with 'secure estab' for example that we send true or false, so this could be stored for any case?

src/test/java/uk/gov/ons/census/casesvc/service/CaseServiceTest.java
```diff
@@ -111,6 +114,115 @@ public void testSaveCaseSample() {
     assertThat(savedCase.getCeActualResponses()).isEqualTo(0);
     assertThat(savedCase.getCaseType()).isEqualTo(TEST_ADDRESS_TYPE);
     assertThat(savedCase.isHandDelivery()).isFalse();
+    assertThat(savedCase.getMetadata().getSecureEstablishment()).isFalse();
+  }
+
+  @Test
+  public void testSaveCaseSampleCE() {
```
LukeLoze1: what does this test add over testSaveCaseSampleCESecureFalse? both test it starts of 0 in the input and ends up being 'saved' as False?

LeoKHoward: Well spotted

PR #124, opened 2020-03-12
Send CLOSE to FWMT for CE U where AR >= ER

NickGrantONS:
# Motivation and Context
When we reach the number of actual responses for a Communal Establishment Unit (e.g. a university halls of residence dorm) then we want to send FWMT a CLOSE instead of an UPDATE.

# What has changed
Created a new bespoke rule which kindly does the needful.

# How to test?
Run the ATs.

Create a CE U case with an expected response count of 2. Link 2 uac-qid pairs to the case. Receipt the QIDs. When the second QID is receipted, a FWMT CLOSE should be sent.

# Links
Trello: https://trello.com/c/GapIaH4L


NickGrantONS:
[Send a close to field when AR>=ER for CE Units (3)](https://trello.com/c/GapIaH4L/697-send-a-close-to-field-when-arer-for-ce-units-3)

PR #125, opened 2020-03-16
Allow Overwriting QID Links

AdamHawtin:
# Motivation and Context
We want to change how we had questionnaire links if the QID in question as already been linked to a different case so that it logs an error but does re-link the QID.

# What has changed
* Log and relink on QID link that has already been linked

# How to test?
Build and run acceptance tests (for regression).
To test the feature, load an unaddressed batch, post in a link for a QID from the batch, then post in a link for the same QID with a different case ID. The case processor should now log and error but continue to relink the QID.

# Links
https://trello.com/c/8VjhtXme/717-overwrite-case-id-if-duplicate-pq-link-qid-event-arrives-5

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
@@ -74,7 +74,7 @@ public class QuestionnaireLinkedService {
       UacDTO uac, UacQidLink uacQidLink, String newLinkedCaseId) {
     if (uacQidLink.getCaze() != null
         && !uacQidLink.getCaze().getCaseId().equals(UUID.fromString(uac.getCaseId()))) {
-      log.with("uac_qid_link_id", uacQidLink.getId())
+      log.with("qid", uacQidLink.getQid())
           .with("previous_case_id", uacQidLink.getCaze().getCaseId())
           .with("new_case_id", newLinkedCaseId)
           .error(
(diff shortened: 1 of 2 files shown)


AdamHawtin:
[Overwrite Case ID if duplicate PQ link QID event arrives (5)](https://trello.com/c/8VjhtXme/717-overwrite-case-id-if-duplicate-pq-link-qid-event-arrives-5)

src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
```diff
@@ -67,11 +70,15 @@ public void processQuestionnaireLinked(
...
   }
 
-  private void checkQidNotLinkedToAnotherCase(UacDTO uac, UacQidLink uacQidLink) {
+  private void logErrorIfQidIsLinkedToAnotherCase(
+      UacDTO uac, UacQidLink uacQidLink, String newLinkedCaseId) {
     if (uacQidLink.getCaze() != null
         && !uacQidLink.getCaze().getCaseId().equals(UUID.fromString(uac.getCaseId()))) {
-      throw new RuntimeException(
-          "UacQidLink already linked to case id: " + uacQidLink.getCaze().getCaseId());
+      log.with("uac_qid_link_id", uacQidLink.getId())
```
NickGrantONS: I'd probably prefer it if we logged the QID - I don't think it's a security risk to have it in the logs. The QID tells us quite a lot without having to access the database

AdamHawtin: Fair point, we do get some important information from the QID itself. 👍

PR #135, opened 2020-04-07
Changing criteria for when an UPDATE message is sent to field

LeoKHoward:
# Motivation and Context
An UPDATE message is still being emitted to field when an undelivered_mail_reported event is ingested
This should not happen if the case is marked as either receipted, refused, invalid or NI CE

# What has changed
New criteria added for when an UPDATE message is not sent
Creation of unit test to cover all scenarios where an UPDATE message is not sent
Had to couple other integration tests to set receipt/refusal/invalid to false

# How to test?
Run and tests locally and check image builds

# Links
https://trello.com/c/hkArUv0E/807-dont-update-field-on-undeliveredmailreported-event-when-case-has-been-receipted-refused-etc-5

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/messaging/UndeliveredMailReceiver.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/messaging/UndeliveredMailReceiver.java
@@ -52,10 +52,7 @@ public class UndeliveredMailReceiver {
               Long.parseLong(event.getPayload().getFulfilmentInformation().getCaseRef()));
     }
 
-    if (!caze.isRefusalReceived()
-        && !caze.isAddressInvalid()
-        && !caze.isReceiptReceived()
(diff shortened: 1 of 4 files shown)


LeoKHoward:
[Don't update field on undelivered_mail_reported event when case has been receipted, refused etc (5)](https://trello.com/c/hkArUv0E/807-dont-update-field-on-undeliveredmailreported-event-when-case-has-been-receipted-refused-etc-5)

src/main/java/uk/gov/ons/census/casesvc/messaging/UndeliveredMailReceiver.java
```diff
@@ -75,4 +77,11 @@ public void receiveMessage(Message<ResponseManagementEvent> message) {
           messageTimestamp);
     }
   }
+
+  private boolean isMarkingCaseAsUpdated(Case caze) {
```
NickGrantONS, line 81: I would call this `shouldEmitCaseUpdated`

PR #138, opened 2020-04-20
Am1 part2 new address using source case

hbrace:
# Motivation and Context

New address events can be submitted that include a case id from an existing case. A number of fields can be copied from either the event or source case, depending whether they are present in the event. There are also a number of fields that do not come on the event, but can either be copied from the source case or not.

# What has changed

New method to build a case from a new address event and source case, taking the fields from the appropriate source depending on presence.
Tests updated and some added.

# How to test?

Build branch, and [test with acceptance tests on the same branch name](https://github.com/ONSdigital/census-rm-acceptance-tests/pull/235). Check the tables in the confluence document linked from Trello, and the confluence doc linked from that doc. Check that these tables match up with how the new case is built.

# Links

https://trello.com/c/3EyvVwQF/708-am1-part-2-new-address-create-using-source-case-8

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
@@ -49,6 +49,28 @@ public class NewAddressReportedService {
         messageTimestamp);
   }
 
+  public void processNewAddressFromSourceId(
+      ResponseManagementEvent newAddressEvent, OffsetDateTime messageTimestamp, UUID sourceCaseId) {
+    CollectionCase newCollectionCase =
+        newAddressEvent.getPayload().getNewAddress().getCollectionCase();
+    checkManadatoryFieldsPresent(newCollectionCase);
+
+    Case sourceCase = caseService.getCaseByCaseId(sourceCaseId);
+    Case newAddressFromSourceCase = buildCaseFromSourceCaseAndEvent(newCollectionCase, sourceCase);
+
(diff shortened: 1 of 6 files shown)


hbrace:
[AM1 part 2 - New Address - create using source case (8)](https://trello.com/c/3EyvVwQF/708-am1-part-2-new-address-create-using-source-case-8)

src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
```diff
@@ -114,4 +137,89 @@ private void checkManadatoryFieldsPresent(CollectionCase newCollectionCase) {
...
+    newCase.setEstabUprn(sourceCase.getEstabUprn());
+    newCase.setAbpCode(sourceCase.getAbpCode());
+    newCase.setHtcDigital(sourceCase.getHtcDigital());
+    newCase.setHtcWillingness(sourceCase.getHtcWillingness());
+    newCase.setLad(sourceCase.getLad());
+    newCase.setLsoa(sourceCase.getLsoa());
+    newCase.setMsoa(sourceCase.getMsoa());
+    newCase.setOa(sourceCase.getOa());
+    newCase.setPrintBatch(sourceCase.getPrintBatch());
+    newCase.setSecretSequenceNumber(sourceCase.getSecretSequenceNumber());
```
NickGrantONS, line 203: Don't copy this over - it's very important that Hibernate sets this because it's what guarantees that our case ref is unique

PR #139, opened 2020-04-20
Case Event for Unlinked QID

LeoKHoward:
# Motivation and Context
Changes show an audit trail for when a QID has been unlinked from a case and linked to a new case

# What has changed
New event type
Removal of warning
Updated test

# How to test?
Run maven clean install along with case-api and then run ATs

# Links
https://trello.com/c/9XSxqY7k/757-create-a-case-event-when-a-qid-has-been-unlinked-from-it-qid-relinking-scenario-5


LeoKHoward:
[Create a case event when a QID has been unlinked from it - QID relinking scenario (5)](https://trello.com/c/9XSxqY7k/757-create-a-case-event-when-a-qid-has-been-unlinked-from-it-qid-relinking-scenario-5)

src/main/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedService.java
```diff
@@ -78,15 +79,24 @@ public void processQuestionnaireLinked(
         messageTimestamp);
   }
 
-  private void logErrorIfQidIsLinkedToAnotherCase(
-      UacDTO uac, UacQidLink uacQidLink, String newLinkedCaseId) {
+  private void logUnlinkedCaseEventAgainstOriginalCaseIfQidIsNowLinkedToNewCase(
+      String questionnaireId,
```
NickGrantONS, line 83: I suppose the questionnaireId is passed in the `uac` object, so you could lose this parameter if you wanted to save a bit of duplication.

src/test/java/uk/gov/ons/census/casesvc/service/QuestionnaireLinkedServiceTest.java
```diff
@@ -342,6 +342,17 @@ public void testQidAlreadyLinkedToDifferentCaseCanBeReLinked() {
             eq(managementEvent.getEvent()),
             anyString(),
             eq(messageTimestamp));
+
+    verify(eventLogger)
+        .logCaseEvent(
+            any(Case.class),
```
NickGrantONS, line 348: it might be tricky to do, but strictly speaking we should check that this is the right case

NickGrantONS: theoretically this should be `testCase`

PR #143, opened 2020-04-23
set new case to skeleton and copy metadata over

hbrace:
# Motivation and Context

If a new address event came in with a source case Id, the new case wasn't being set as a skeleton case. Also, the case metadata needed copying over (at the moment, only secure establishment).

# What has changed

New case has its skeleton flag ticked.
Method added to set the case metadata (might be useful for future if more properties are added to the case metadata. If there is something in the case metadata we don't want being copied into the new case, we can set here).
Tests updated.

# How to test?

maven clean install, run alongside acceptance tests on branch of [the same name](https://github.com/ONSdigital/census-rm-acceptance-tests/pull/239).

# Links

https://trello.com/c/2pT2lf1x/874-add-more-fields-from-source-case-id-upon-receiving-newaddressreported-3

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
@@ -14,11 +14,13 @@ import uk.gov.ons.census.casesvc.model.dto.EventTypeDTO;
 import uk.gov.ons.census.casesvc.model.dto.Metadata;
 import uk.gov.ons.census.casesvc.model.dto.ResponseManagementEvent;
 import uk.gov.ons.census.casesvc.model.entity.Case;
+import uk.gov.ons.census.casesvc.model.entity.CaseMetadata;
 import uk.gov.ons.census.casesvc.model.entity.EventType;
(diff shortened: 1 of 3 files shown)


hbrace:
[Add more fields from source case id upon receiving NewAddressReported (3)](https://trello.com/c/2pT2lf1x/874-add-more-fields-from-source-case-id-upon-receiving-newaddressreported-3)

src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
```diff
@@ -227,9 +237,11 @@ private Case buildCaseFromSourceCaseAndEvent(CollectionCase newCollectionCase, C
     newCase.setOa(sourceCase.getOa());
     newCase.setPrintBatch(sourceCase.getPrintBatch());
     newCase.setSurvey(sourceCase.getSurvey());
+    newCase.setMetadata(metadataFromSourceCase(sourceCase.getMetadata()));
```
NickGrantONS, line 240: I thought we agreed we weren't going to do this - it should be explicitly copying the `secureEstablishment` value into a brand new `CaseMetadata` object.

PR #145, opened 2020-05-05
910 copy lat long from source case id overwrite with anything on event

LukeLoze1:
# Motivation and Context
When a newAddress reported has no Long/Lat on the event, if it has a source case copy them from that

# What has changed
Copies source rather than null for lat/long

# How to test?
mvn clean install
& with ATs on card

# Links
https://trello.com/c/nPtXCwKt/910-copy-lat-long-from-source-case-id-overwrite-with-anything-on-event-3

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
@@ -215,9 +215,12 @@ public class NewAddressReportedService {
     // Set fields empty/null/blank unless they come from the event
     newCase.setOrganisationName(
         getEventValOverSource(null, newCollectionCase.getAddress().getOrganisationName()));
-    newCase.setLatitude(getEventValOverSource(null, newCollectionCase.getAddress().getLatitude()));
+    newCase.setLatitude(
+        getEventValOverSource(
+            sourceCase.getLatitude(), newCollectionCase.getAddress().getLatitude()));
     newCase.setLongitude(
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
```diff
@@ -215,9 +215,12 @@ private Case buildCaseFromSourceCaseAndEvent(CollectionCase newCollectionCase, C
     // Set fields empty/null/blank unless they come from the event
     newCase.setOrganisationName(
         getEventValOverSource(null, newCollectionCase.getAddress().getOrganisationName()));
-    newCase.setLatitude(getEventValOverSource(null, newCollectionCase.getAddress().getLatitude()));
+    newCase.setLatitude(
```
NickGrantONS, line 218: please move this into the section above

src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
```diff
@@ -215,9 +215,12 @@ private Case buildCaseFromSourceCaseAndEvent(CollectionCase newCollectionCase, C
...
+        getEventValOverSource(
+            sourceCase.getLatitude(), newCollectionCase.getAddress().getLatitude()));
     newCase.setLongitude(
-        getEventValOverSource(null, newCollectionCase.getAddress().getLongitude()));
+        getEventValOverSource(
```
NickGrantONS, line 222: please move this into the section above

PR #147, opened 2020-05-11
Don't unreceipt cases which were receipted by EQ

NickGrantONS:
# Motivation and Context
We want to stop the unreceipting of the tiny handful of cases which have been receipted via EQ, but the respondent also posted back the blank questionnaire.

# What has changed
Check to see if an EQ receipt has been received on the same QID and ignore the unreceipt if it has.

# How to test?
Load a sample. Receipt the case via EQ. Then, receipt it again via the paper receipting process, but mark it as blank. The blank should be ignored.

# Links
Trello: https://trello.com/c/yAgPc1W1

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/QidReceiptService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/QidReceiptService.java
@@ -10,6 +10,7 @@ import uk.gov.ons.census.casesvc.logging.EventLogger;
 import uk.gov.ons.census.casesvc.model.dto.ResponseDTO;
 import uk.gov.ons.census.casesvc.model.dto.ResponseManagementEvent;
 import uk.gov.ons.census.casesvc.model.entity.Case;
+import uk.gov.ons.census.casesvc.model.entity.Event;
 import uk.gov.ons.census.casesvc.model.entity.EventType;
 import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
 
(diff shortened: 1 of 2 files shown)


NickGrantONS:
[Spike - Blank PQ double receipt scenario (1 day)](https://trello.com/c/yAgPc1W1/865-spike-blank-pq-double-receipt-scenario-1-day)

src/main/java/uk/gov/ons/census/casesvc/service/QidReceiptService.java
```diff
@@ -72,6 +73,15 @@ public void processUnreceipt(
       ResponseManagementEvent unreceiptEvent,
       OffsetDateTime messageTimestamp,
       UacQidLink uacQidLink) {
+    if (!uacQidLink.isActive() && hasEqReceipt(uacQidLink)) {
```
davidmort, line 76: do we need to check for an eq receipt that has come via any UacQid links not just this one?

NickGrantONS: It says "Blank Questionnaire received and valid receipt received with same QID" on the ticket, but it's a possible scenario, sure.

AdamHawtin: Blank questionnaires should already be logged and ignored if another QID has been receipted so I think this scenario only applies to receipts for the same QID?

davidmort: yeah that's a fair point

AdamHawtin:
Changes rolled into #152

PR #149, opened 2020-05-13
AM1 7.1 new address HI telephone capture

AdamHawtin:
# Motivation and Context
Child cases from skeleton cases should be marked as such

# What has changed
* Copy over skeleton marker when creating a child individual case

# How to test?
Run the new AT for individual telephone capture on newly reported address, check the skeleton flag is set in the DB on the HI case created

# Links
https://trello.com/c/zwSsVjsD/713-am1-part-71-new-address-create-hi-case-for-telephone-capture-3
https://github.com/ONSdigital/census-rm-acceptance-tests/pull/249


AdamHawtin:
[AM1 part 7.1 - New Address - Create HI Case for telephone capture (3)](https://trello.com/c/zwSsVjsD/713-am1-part-71-new-address-create-hi-case-for-telephone-capture-3)

PR #151, opened 2020-05-14
Add missing indexes to improve performance

NickGrantONS:
# Motivation and Context
Case API was performing very badly when retrieving case events.

# What has changed
Added indexes to both `uac_qid_link` and `event` table to prevent table-scanning.

# How to test?
Load large sample. Call case API including `?caseEvents=true`. It should be fast response time.

# Links
Trello: https://trello.com/c/fHPtdl59


NickGrantONS:
[Case API performance is horrible when retrieving case events](https://trello.com/c/fHPtdl59/937-case-api-performance-is-horrible-when-retrieving-case-events)

PR #152, opened 2020-05-18
Handle EQ receipt and blank questionnaire scenarios

AdamHawtin:
# Motivation and Context
EQ receipts should trump blank questionnaires since and always receipt the case regardless of blank questionnaires.

# What has changed
* Don't ignore eQ receipts on a blank marked QID 
* Don't unreceipt eQ receipts (adapted from #147)

# How to test?
Run the linked AT's branch

# Links
https://trello.com/c/Uwfnumcg/953-blank-pq-eq-receipt-5
https://github.com/ONSdigital/census-rm-acceptance-tests/pull/255

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceiptService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceiptService.java
@@ -13,6 +13,7 @@ import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import org.springframework.stereotype.Component;
 import uk.gov.ons.census.casesvc.model.dto.ActionInstructionType;
+import uk.gov.ons.census.casesvc.model.dto.EventDTO;
 import uk.gov.ons.census.casesvc.model.dto.EventTypeDTO;
 import uk.gov.ons.census.casesvc.model.entity.Case;
 import uk.gov.ons.census.casesvc.model.entity.UacQidLink;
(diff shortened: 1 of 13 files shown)


AdamHawtin:
[Blank PQ EQ receipt (5)](https://trello.com/c/Uwfnumcg/953-blank-pq-eq-receipt-5)

src/main/java/uk/gov/ons/census/casesvc/service/CaseReceiptService.java
```diff
@@ -53,10 +55,12 @@ private void setUpRules() {
...
   }
 
-  public void receiptCase(UacQidLink uacQidLink, EventTypeDTO causeEventType) {
+  public void receiptCase(UacQidLink uacQidLink, EventDTO causeEvent) {
     Case caze = uacQidLink.getCaze();
 
-    if (caze.isReceiptReceived() || uacQidLink.isBlankQuestionnaire()) return;
+    if (caze.isReceiptReceived()
+        || (uacQidLink.isBlankQuestionnaire() && !causeEvent.getChannel().equals(EQ_EVENT_CHANNEL)))
+      return;
```
NickGrantONS, line 63: I was thinking about this, and I think we want to record the case event against the UAC-QID, even if we don't do anything else.

AdamHawtin:
> I think we should record the case event against the UAC-QID, even if we don't do anything else (like emitting a rabbit message).

That should be the current beheviour, the `CaseReceiptService` is called by the `QidReceiptService` and only handles actions against the case. All the events should still get logged by the `QidReceiptService`.

PR #153, opened 2020-05-20
Handle Fulfilment Confirmed events

NickGrantONS:
# Motivation and Context
We want to log when a fulfilment has been confirmed as printed.

# What has changed
Added message listener endpoint to log fulfilment confirmed events.

# How to test?
Drop messages onto the fulfilment-confirmed PubSub topic and then check the case events in the DB.

# Links
Trello: https://trello.com/c/QShkkGg6


NickGrantONS:
[RMC-203 Handle Fulfilment Confirmed event (13)](https://trello.com/c/QShkkGg6/618-rmc-203-handle-fulfilment-confirmed-event-13)

PR #154, opened 2020-05-21
943 fulfilment mappings ind response print products

ryangrundy7:
# Motivation and Context

Added fulfilment codes for the new UAC Individual print letters. For these new codes it'll generate a new HI case from it.

# What has changed

- Added new fulfilment codes to individual response set

# How to test?

- Run mvn clean install
- Run with ATs with other branches

# Links

[Trello](https://trello.com/c/2pVJow0S)


ryangrundy7:
[Fulfilment mappings - Ind Response Print Products (5)](https://trello.com/c/2pVJow0S/943-fulfilment-mappings-ind-response-print-products-5)

PR #155, opened 2020-05-22
Bump Spring version

NickGrantONS:
# Motivation and Context
We should keep Spring Boot up to date, because newer versions contain many security, stability and performance improvements, and it ensures that we never end up running an unsupported old version

# What has changed
Bumped Spring up to v2.2.7.

# How to test?
Should be zero regression.

# Links
Trello: https://trello.com/c/sNmbnD6B


NickGrantONS:
[Bump Spring Boot version (5)](https://trello.com/c/sNmbnD6B/955-bump-spring-boot-version-5)

src/main/java/uk/gov/ons/census/casesvc/config/MessageConsumerConfig.java
```diff
@@ -364,4 +343,26 @@ private SimpleMessageListenerContainer setupListenerContainer(
...
+          public Map<String, Object> toHeadersFromRequest(MessageProperties source) {
+            Map<String, Object> headers = new HashMap<>();
+
+            /* We strip EVERYTHING out of the headers and put the content type back in because we
+             * don't want the __TYPE__ to be processed by Spring Boot, which would cause
+             * ClassNotFoundException because the type which was sent doesn't match the type we
+             * want to receive. This is an ugly workaround, but it works.
+             */
+            headers.put("contentType", "application/json");
+            return headers;
```
AdamHawtin, line 363: Can this be tested? Even with the comment I'm not exactly sure what problem this is solving.

NickGrantONS: If you take it out and run the acceptance tests then you'll see the problem - Spring services are sending Rabbit messages which include a `__TYPEID__` header, containing the fully-qualified class name and package of the DTO object, which obviously doesn't work if the message comes from RH/CC/FWMT or in-between different parts of RM (e.g. Case API -> Case Processor, or Case Processor -> Action Scheduler). The new version of Spring Boot that we're using is more strict than the old version, and it insists on using the `__TYPEID__` if it's present in the headers.

The alternative solution would be to remove it every where, but that would require code changes in RH, CC and FWMT, which all currently send the dreaded `__TYPEID__` header.

AdamHawtin: 👍 Thanks for clarifying

PR #160, opened 2020-06-05
Fix flakey slow integration tests

NickGrantONS:
# Motivation and Context
The case processor integration tests were slowing down the build more and more, with it taking in excess of 7 minutes. Also, lots of tests had to 'dirty' the Spring context in order to work reliably, which was an ugly hack.

# What has changed
Handle the cleanup of temporary queue listeners neatly and elegantly, such that tests which listen to the same queue don't end up consuming other tests' messages.

# How to test?
Should be like-for-like... only QUICKER!

# Links
Trello: https://trello.com/c/zrCqMG4A


NickGrantONS:
[Sort out the flakey, slow, dodgy Case Processor integration tests](https://trello.com/c/zrCqMG4A/1016-sort-out-the-flakey-slow-dodgy-case-processor-integration-tests)

PR #163, opened 2020-06-11
link questionnaire to CE/SPG parent case even if ind case Id is present

hbrace:
# Motivation and Context

Currently if a questionnaire linked event comes in with an individual case id present and the parent case is CE/SPG, it'll chuck an error. We now want to link the questionnaire to the parent case and ignore the individual case id.

# What has changed

Error chucking stuff removed. Tests updated.

# How to test?

Build the branch. Test the branch with the [acceptance tests on the branch of the same name.](https://github.com/ONSdigital/census-rm-acceptance-tests/pull/273)
# Links

https://trello.com/c/V7O2qqSV/1017-pq-link-ignore-individual-case-id-if-parent-case-is-not-hh-5


hbrace:
[PQ link - Ignore Individual Case ID if 'parent' case is not HH? (5)](https://trello.com/c/V7O2qqSV/1017-pq-link-ignore-individual-case-id-if-parent-case-is-not-hh-5)

PR #165, opened 2020-06-12
surveyLaunched set to true is survey launched from RH

LukeLoze1:
# Motivation and Context
New Survey Launched flag and handling

# How to test?
mvn clean install and with other apps from this ticket

# Links
https://trello.com/c/jRDDwnf8/981-reminders-started-eq-but-not-submitted-13


src/main/java/uk/gov/ons/census/casesvc/model/dto/CollectionCase.java
```diff
@@ -37,4 +37,5 @@
   private boolean skeleton;
   private CaseMetadata metadata;
   private String printBatch;
+  private Boolean surveyLaunched;
```
AdamHawtin, line 40: Could be `boolean`?

src/main/java/uk/gov/ons/census/casesvc/service/SurveyService.java
```diff
@@ -30,6 +33,13 @@ public void processMessage(ResponseManagementEvent surveyEvent, OffsetDateTime m
     UacQidLink surveyLaunchedForQid =
         uacService.findByQid(surveyEvent.getPayload().getResponse().getQuestionnaireId());
 
+    Case caze = surveyLaunchedForQid.getCaze();
+
+    if (caze != null && surveyEvent.getEvent().getChannel().equals("RH")) {
```
AdamHawtin, line 38: I know we're only expecting these events from RH, but should we be effectively ignoring them if they come from a different channel?

LukeLoze1: It's not ignoring it, it just won't update the column - because we only want to update the column if it's linked to a case & it's from RH.

This was why I wanted the column called something like rh_survey_launched.

PR #167, opened 2020-06-16
tidy blank questionnaire table to be more readable

LukeLoze1:
# Motivation and Context
It was needlessly verbose before

# What has changed
Remove 10 zillion needless 'BlankQuestionnaireService.' prefixes and some annotations from the test 

# How to test?
mvn clean install

### DIFF
--- a/src/test/java/uk/gov/ons/census/casesvc/service/BlankQuestionnaireServiceTest.java
+++ b/src/test/java/uk/gov/ons/census/casesvc/service/BlankQuestionnaireServiceTest.java
@@ -15,7 +15,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.UUID;
 import lombok.AllArgsConstructor;
-import lombok.EqualsAndHashCode;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/ons/census/casesvc/service/BlankQuestionnaireServiceTest.java
```diff
@@ -384,7 +383,6 @@ private String getQid(String qidType) {
   }
 
   @AllArgsConstructor
-  @EqualsAndHashCode
```
NickGrantONS, line 387: How is your map going to work without equals and hashcode on the key?

NickGrantONS: The fact that you can remove this indicates that the unit tests are broken/flawed

LukeLoze1: There's no Map in the unit tests.
The Key class is a data class to set up the parameters for the tests.

PR #170, opened 2020-06-19
Log useful stacktrace line

NickGrantONS:
# Motivation and Context
When we get errors like null pointer exceptions, it's often unclear where exactly in the code things have gone wrong. The stacktrace has the detail, but we don't log stacktraces by default, and stack traces are often laborious to trawl through.

# What has changed
Added code to trawl the stacktrace and extract the 'root cause' in our code of where the NPE or other type of exception originated from, which should allow us to more quickly diagnose & fix problems.

# How to test?
Send in a bad message (missing some mandatory data, for example) and check the logs - there should be details about the line of code which threw the exception (NOT the cause... the RESULT of bad data being sent into RM which shouldn't happen but does happen when we're testing).

# Links
Trello: https://trello.com/c/DQgYqRA8

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/client/ExceptionManagerClient.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/client/ExceptionManagerClient.java
@@ -23,11 +23,16 @@ public class ExceptionManagerClient {
   private String port;
 
   public ExceptionReportResponse reportException(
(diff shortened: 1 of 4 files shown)


NickGrantONS:
[Log exception root cause to aid diagnostics of NPEs etc](https://trello.com/c/DQgYqRA8/1081-log-exception-root-cause-to-aid-diagnostics-of-npes-etc)

PR #172, opened 2020-06-24
Check ceExpectedCapacity not null when comparing actual to expected for CE U

hbrace:
# Motivation and Context

There's a possibility that ceExpectedCapacity could be null. When a receipt for a CE Unit is received the actual response is incremented and then compared to ceExpectedCapacity. If ceExpectedCapacity is null a null pointer exception was thrown.
# What has changed

Added null check to comparison and moved into separate method.
# How to test?

The acceptance test receipted cases increment ceActualResponses is useful for this. Have a break point set after the sample is loaded for the CE U test and set the ceExpectedCapacity to null in the DB. Let it run and it shouldn't null pointer exception.
# Links

https://trello.com/c/EtgoTcNJ/1077-incrementing-actual-response-count-is-broken-if-expected-response-count-is-null

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceiptService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/CaseReceiptService.java
@@ -141,8 +141,7 @@ public class CaseReceiptService {
 
         Case lockedCase = incrementNoReceipt(caze);
 
-        if (!caze.isReceiptReceived()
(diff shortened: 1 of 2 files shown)


hbrace:
[Incrementing actual response count is broken if expected response count is null](https://trello.com/c/EtgoTcNJ/1077-incrementing-actual-response-count-is-broken-if-expected-response-count-is-null)

PR #174, opened 2020-07-03
Spike sending PubSub message to AIMS for new addresses

NickGrantONS:
# Motivation and Context
Fieldworkers might discover new addresses, and they want to include them, but they need to have an entry in AIMS or else RH won't be able to launch. Case Processor needs to tell AIMS about new addresses, instead of having AIMS subscribe to the relevant Rabbit topic.

# What has changed
Added pubsub to Case Processor and got it to send messages to a topic in a project, which can. be configured to be in AIMS.

# How to test?
There are integration tests. If you want to run manually, you can point at your own GCP project by fiddling with the `application.yml`. Make sure you have run `gcloud auth application-default login` and set your `GOOGLE_APPLICATION_CREDENTIALS`. Basically, it's pretty fussy about credentials, but it will work with some hacking. You'll need a PubSub topic to publish to in your project, if you're not using the emulator.

# Links
Trello: https://trello.com/c/oBRB9Vtc


NickGrantONS:
[Send dummy UPRN to AIMS (3 days)](https://trello.com/c/oBRB9Vtc/1051-send-dummy-uprn-to-aims-3-days)

src/main/resources/application.yml
```diff
@@ -94,4 +99,8 @@ censusconfig:
 uprnconfig:
   dummyuprnprefix: 999
 
+pubsub:
+  publishtimeout: 30  # In seconds
+  aimstopic: aims-topic
```
AdamHawtin, line 104: Worth naming this something like `new-address-topic` to clarify it's purpose?

NickGrantONS: not a bad shout

PR #175, opened 2020-07-07
Change refusal for estab case condition

liamtoozer:
# Motivation and Context

All refusals should now be sent to Fieldwork unless the original message channel is from Fieldwork.

# What has changed

* Remove logic bits for checking "estab case and channel not from field"
* Updated test to check null metadata is sent in outbound message to fieldwork-adapter (which should then ignore the message and not send to Fieldwork)

# How to test?

Run the acceptance tests

# Links

https://trello.com/c/u85DtO8j/1045-refusal-table-of-truth-changes-3


liamtoozer:
[Refusal table of truth changes (3)](https://trello.com/c/u85DtO8j/1045-refusal-table-of-truth-changes-3)

PR #180, opened 2020-07-23
Address Modified Event

LeoKHoward:
# Motivation and Context
Need changes for an Address Modified Event to be reflected rather than just logged which is what is currently happening
 
# What has changed
New Address Modified DTO
New Address Modified Service file listing estab types and fields that can be modified
Integration and unit tests for happy path and also checking it fails if mandatory field is missing

# How to test?
Maven clean install in repo, check tests pass and image builds then run ATs

# Links
https://trello.com/c/tgkYwhu1/1186-am-event-modify-8
https://collaborate2.ons.gov.uk/confluence/display/CRM/ESTAB_TYPES+2021
https://collaborate2.ons.gov.uk/confluence/display/CATD/Response+Management+Event+Dictionary


LeoKHoward:
[AM - Event - Modify (8)](https://trello.com/c/tgkYwhu1/1186-am-event-modify-8)

src/test/java/uk/gov/ons/census/casesvc/messaging/AddressReceiverIT.java
```diff
@@ -149,16 +149,114 @@ public void testInvalidAddress() throws Exception {
...
+      managementEvent
+          .getPayload()
+          .getAddressModification()
+          .getNewAddress()
+          .setOrganisationName("modified org name");
+      managementEvent
+          .getPayload()
+          .getAddressModification()
+          .getNewAddress()
+          .setEstabType("HOSPITAL");
```
NickGrantONS, line 211: These billions of lines could be reduced to 6 by pulling out the `newAddress` object

src/main/java/uk/gov/ons/census/casesvc/service/AddressModificationService.java
```diff
@@ -0,0 +1,107 @@
...
+    validate(addressModification);
+
+    Case caze = caseService.getCaseByCaseId(addressModification.getCollectionCase().getId());
+
+    caze.setEstabType(addressModification.getNewAddress().getEstabType());
+    caze.setTownName(addressModification.getNewAddress().getTownName());
+    caze.setAddressLine1(addressModification.getNewAddress().getAddressLine1());
+    caze.setAddressLine2(addressModification.getNewAddress().getAddressLine2());
+    caze.setAddressLine3(addressModification.getNewAddress().getAddressLine3());
+    caze.setOrganisationName(addressModification.getNewAddress().getOrganisationName());
```
AdamHawtin, line 82: Would this cope with being able to modify only a subset of the fields? It looks like if they send us a message with, for example, a missing or `null` organisation name this would null it on the case, I don't see a way to leave these fields out of modification

NickGrantONS: we might be able to use `Optional` in the DTO... gonna do an experiment

NickGrantONS: have a look now. I don't think it implements the requirement (process address modification messages) any better than it did before, because CC/FWMT-G are not going to send us hand-crafted JSON with only the unique bits that have changed - they're just going to splat the whole address because that's easiest

AdamHawtin: If we can guaranty we're always going to be sent everything then I'd agree it's unnecessary, but since the purpose of this is vanity changes I was under the impression they would be trying to update minimal fields. I also thought we were avoiding asking people to supply the current, unchanged data because that opens the possibility of accidentally reverting other changes. 

Do we need to clarify exactly how this is intended to be used before proceeding with this? The example in the event dictionary still has an `oldAddress` block, but does also leave `townName` out of the `newAddress` so I'm not sure what to make of it.

src/main/resources/application.yml
```diff
@@ -88,3 +88,45 @@ censusconfig:
 
 uprnconfig:
   dummyuprnprefix: 999
+
+approvedlistofgovernmentauthorisedpremises: >
```
AdamHawtin, line 92: a name mentioning `estabType` would be more helpful

PR #181, opened 2020-07-27
1183 am file new

ryangrundy7:
# Motivation and Context

When adding new addresses from AR as new cases in RM, we don't want them to create uac-qids and we want to send a create message to field when this has happened.

# What has changed

- When new addresses come in from the bulk processor, it creates cases without the uac-qids attached to the case.

# How to test?

- Run with other branches on PR ([toolbox](https://github.com/ONSdigital/census-rm-toolbox/pull/56), terraform and kubernetes). Should be able to put a file in the bucket and once the bulk processor has ran, new cases should be created from the addresses.

# Links

[Trello](https://trello.com/c/ATLrBzMk)
[Toolbox](https://github.com/ONSdigital/census-rm-toolbox/pull/56)
[Kubernetes](https://github.com/ONSdigital/census-rm-kubernetes/pull/288)
[Terraform](https://github.com/ONSdigital/census-rm-terraform/pull/186)

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/EventService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/EventService.java
@@ -20,9 +20,9 @@ import uk.gov.ons.census.casesvc.utility.QuestionnaireTypeHelper;
 public class EventService {
 
   public static final String CREATE_CASE_SAMPLE_RECEIVED = "Create case sample received";
-  public static final String CREATE_BULK_CASE_SAMPLE_RECEIVED = "Create bulk case sample received";
-  public static final String BULK_PROCESSING_EVENT_CHANNEL = "AR";
-  public static final String BULK_PROCESSING_EVENT_SOURCE = "RM";
+  private static final String CREATE_BULK_CASE_SAMPLE_RECEIVED = "Create bulk case sample received";
+  private static final String BULK_PROCESSING_EVENT_CHANNEL = "AR";
+  private static final String BULK_PROCESSING_EVENT_SOURCE = "RM";
 
   private final CaseService caseService;
   private final UacService uacService;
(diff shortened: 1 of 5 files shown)


ryangrundy7:
[AM - File - New (13)](https://trello.com/c/ATLrBzMk/1183-am-file-new-13)

src/main/java/uk/gov/ons/census/casesvc/service/EventService.java
```diff
@@ -35,22 +40,39 @@ public void processSampleReceivedMessage(
...
+              BULK_PROCESSING_EVENT_CHANNEL,
+              BULK_PROCESSING_EVENT_SOURCE),
+          convertObjectToJson(createCaseSample),
+          messageTimestamp);
+    } else {
+      UacQidLink uacQidLink = uacService.buildUacQidLink(caze, questionnaireType);
       uacService.saveAndEmitUacUpdatedEvent(uacQidLink);
+      caseService.saveCaseAndEmitCaseCreatedEvent(caze);
+
+      eventLogger.logCaseEvent(
```
NickGrantONS, line 63: feels like this code repetition was easily avoidable (i.e. just varying the event description isn't hard)

src/main/java/uk/gov/ons/census/casesvc/service/EventService.java
```diff
@@ -18,6 +20,9 @@
 public class EventService {
 
   public static final String CREATE_CASE_SAMPLE_RECEIVED = "Create case sample received";
+  public static final String CREATE_BULK_CASE_SAMPLE_RECEIVED = "Create bulk case sample received";
+  public static final String BULK_PROCESSING_EVENT_CHANNEL = "AR";
+  public static final String BULK_PROCESSING_EVENT_SOURCE = "RM";
```
NickGrantONS, line 25: these things are only public for the convenience of the tests, which is very bad practice. Should be private

NickGrantONS: they are just constants though so I guess it's not the end of the world, but it's a bad habit to get into

liamtoozer, line 25: Might be also worth stating what `AR` stands for in the name as well?

PR #185, opened 2020-07-30
Hack around the problem with retry messages

NickGrantONS:
# Motivation and Context
An acceptance test was intermittently failing, indicating a problem with RM.

# What has changed
Tweak UAC cache timeout to stop getting errors. Reduced retry count to 1 for failed rabbit messages, because that was causing messages to be processed more than once.

# How to test?
Run the scenario `Tranche 2 household case details to be sent to the Field Work Management Tool`. To simulate the problem seen in CI, increase the UAC cache fetch count to 3,000.

# Links
Trello: https://trello.com/c/SCSNxVxU


NickGrantONS:
[ATs fail 5 out of 8 times in CI](https://trello.com/c/SCSNxVxU/1246-ats-fail-5-out-of-8-times-in-ci)

PR #186, opened 2020-08-03
Increase retry delay for eventual consistency

NickGrantONS:
# Motivation and Context
ATs are failing because Notify Processor is too quick and it sends the RM_UAC_CREATED message before the case has been created.

# What has changed
Increased the retry delay from 1 second to 3 seconds.

# How to test?
Run the ATs - they'll probably work locally though - it's only broken in CI.

# Links
Trello: https://trello.com/c/RExzTzH3


NickGrantONS:
[AT nightmare](https://trello.com/c/RExzTzH3/1265-at-nightmare)

PR #189, opened 2020-08-06
Fix Spring retries bug with transaction manager

NickGrantONS:
# Motivation and Context
If we hit a problem when processing a Rabbit message, we aren't rolling back and handling retries properly - we are sending extra (duplicate) Rabbit messages, amongst other problems. This is quite fundamental to the reliable functioning of RM, so we can trust our code not to corrupt any data in the event of exceptions.

# What has changed
Added a `ChainedTransactionManager` per the suggestion of Dr David Syer in his article _Distributed transactions in Spring, with and without XA_ (http://www.infoworld.com/article/2077963/distributed-transactions-in-spring--with-and-without-xa.html)

# How to test?
It's super hard.

Suggestion: set the cache fetch count to 3,000 or more, and the timeout to 5 seconds or less, then run the AT scenario `Tranche 2 household case details to be sent to the Field Work Management Tool` with a freshly booted up Case Processor - you would see duplicate cases being emitted, erroneously, in the original code.

# Links
Trello: https://trello.com/c/J4uL5FRY

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/config/AppConfig.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/config/AppConfig.java
@@ -2,7 +2,6 @@ package uk.gov.ons.census.casesvc.config;
 
 import java.util.TimeZone;
 import javax.annotation.PostConstruct;
-import javax.persistence.EntityManagerFactory;
 import ma.glasnost.orika.MapperFacade;
 import ma.glasnost.orika.MapperFactory;
 import ma.glasnost.orika.impl.DefaultMapperFactory;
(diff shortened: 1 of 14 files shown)


NickGrantONS:
[Spring retries - what the heck?](https://trello.com/c/J4uL5FRY/1273-spring-retries-what-the-heck)

NickGrantONS:
> this is great work - do you think the sampleReceiver integration test needs updating like the others?

I'm satisfied that we have enough test coverage to stop regression, and protecting the retries and transactional stuff that was causing us problems - no reason to do anything more to the sample ITs IMHO

davidmort:
> > this is great work - do you think the sampleReceiver integration test needs updating like the others?
> 
> I'm satisfied that we have enough test coverage to stop regression, and protecting the retries and transactional stuff that was causing us problems - no reason to do anything more to the sample ITs IMHO

yeah thats fair enough, was kinda thinking that myself bu thought id ask the question 👍

PR #193, opened 2020-08-18
Allow batch ID to be sent on RM_UAC_CREATED

NickGrantONS:
# Motivation and Context
Although we don't store batch IDs anywhere - they're transient - we will be able to see all the UACs created by a particular Action (rule or fulfilment) batch, grouped together.

# What has changed
Allow batch ID to be sent, received and stored against any UACs created by RM.

# How to test?
Trigger a reminder rule. All the UACs that get created should have the same batch ID.

# Links
Trello: https://trello.com/c/ELepDger


NickGrantONS:
[Use batch ID on UAC QID links for scheduled actions (5)](https://trello.com/c/ELepDger/261-use-batch-id-on-uac-qid-links-for-scheduled-actions-5)

PR #206, opened 2020-09-16
1406 Address type change address modification

AdamHawtin:
# Motivation and Context
It has been decided that we need to accept address modifications within the address type change event.

# What has changed
* Add address type change address DTO object
* Allow modifying certain address fields in the address type change event
* Tests

# How to test?
Run with branch AT's

# Links
https://trello.com/c/wMcpQM1K/1406-allow-address-modifications-on-an-address-type-changed-event-from-cc-13


AdamHawtin:
[Allow address modifications on an address type changed event from CC (13)](https://trello.com/c/wMcpQM1K/1406-allow-address-modifications-on-an-address-type-changed-event-from-cc-13)

src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
```diff
@@ -132,4 +186,26 @@ private String deriveAddressLevel(String oldCaseType, String newCaseType) {
       throw new RuntimeException("Invalid Case Type change");
     }
   }
+
+  private void validateModification(AddressTypeChangeAddress address) {
```
NickGrantONS, line 190: This method looks like a carbon copy of `AddressModificationService.validate()`.

Given that it's only checking two values, would it be possible to turn it into a helper method, to avoid code duplication? Not really worth me blocking the PR for though

NickGrantONS (approved):
Just a minor comment about code re-use, but otherwise looks great

src/main/java/uk/gov/ons/census/casesvc/utility/AddressModificationValidator.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.ons.census.casesvc.utility;
+
+import java.util.Set;
+import org.springframework.util.StringUtils;
+import uk.gov.ons.census.casesvc.model.dto.ModifiedAddress;
+
+public class AddressModificationValidator {
```
NickGrantONS, line 7: hmmm. I would have done this as a `@Component` and had the `estabTypes` injected into it by Spring

NickGrantONS (changes requested):
Nice refactoring, but having done it like that, can I request that the estab types are injected into `AddressModificationValidator` `@Component` to cut down on the amount of places we have to inject that set, and pass it around

src/main/java/uk/gov/ons/census/casesvc/model/dto/ModifiedAddress.java
```diff
@@ -6,11 +6,12 @@
 import lombok.Data;
 
 @Data
-@JsonInclude(Include.NON_ABSENT) // Only needed for marshalling (i.e. sending)
+@JsonInclude(Include.NON_NULL) // Only needed for marshalling (i.e. sending)
```
NickGrantONS, line 9: I think `NON_ABSENT` was the right value, meaning "if it's not present, don't put it in the JSON" - that'll be needed for what we log, so that we know what was sent to us in the event

NickGrantONS: you can put `NON_ABSENT` onto all the optionals, and `NON_NULL` onto the addressType maybe?

src/main/java/uk/gov/ons/census/casesvc/utility/AddressModificationValidator.java
NickGrantONS: but then I suppose we'd have to inject the `AddressModificationValidator` into the two places it's needed.

Hmmm. Disregard if you want

src/main/java/uk/gov/ons/census/casesvc/model/dto/ModifiedAddress.java
AdamHawtin: I only saw the correct behaviour in the integration tests with `NON_NULL`. 

`NON_ABSENT` seems to not include empty optionals.

NickGrantONS: I'm sure that `NON_ABSENT` is what we want - if you want your IT not to send a value, you might have to try with it null'ed. Let's chat on Slack about this

NickGrantONS: Fundamentally, we shouldn't change the production code for the sake of the tests. If we need the tests to do something weird/different, we'll have to find a different way to build the messages. The most important thing is that the JSON payload persisted to the event log, matches the JSON that we received.

AdamHawtin: As I understand it, that annotation is only there for the tests at the moment anyway? Our production code never marshals into this DTO. The problem I had with `NON_ABSENT` is it ignores both `null` values and `Optional.empty` values, giving me no way to distinguish between the json values being `null` and being not present in the tests. Happy to have a short conversation, it might be me doing the marshalling wrong.

NickGrantONS: I was wrong! That'll learn me for not testing this for myself!

liamtoozer (approved):

NickGrantONS (approved):
Love that `JsonHelperTest`

PR #207, opened 2020-09-16
new CLERICAL_ADDRESS_RESOLUTION eventType for bulk new Address

LukeLoze1:
# Motivation and Context
Mark and record bulk address resolution events as snappy clerical address resolution events.  This will be used by fieldworkadapter 

# What has changed
New Type

Test by itself and with other from ticket


LeoKHoward (approved):
Works for me

NickGrantONS (approved):
Nice

PR #209, opened 2020-09-24
Consume Secure Flag On New Address Reported

LeoKHoward:
# Motivation and Context
We want to record a property as secure so that future visits can be arranged if required

# What has changed
New boolean secureType in Address DTO...can only be true or false (not null)
Checks if case metadata exists and if so what the secureEstablishment flag is set to
Test

# How to test?
Build branch and run ATs

# Links
https://trello.com/c/4fnNI9lZ/1231-consume-secure-flag-on-newaddressreported-5
https://collaborate2.ons.gov.uk/confluence/pages/viewpage.action?pageId=40216113


LeoKHoward:
[Consume secure flag on newAddressReported (5)](https://trello.com/c/4fnNI9lZ/1231-consume-secure-flag-on-newaddressreported-5)

src/main/java/uk/gov/ons/census/casesvc/model/dto/Address.java
```diff
@@ -19,4 +19,5 @@
   private String addressLevel;
   private String estabType;
   private String organisationName;
+  private boolean secureType;
```
NickGrantONS, line 22: I've been testing... it seems like this is fatal if there's no `secureType` supplied in the JSON, or a null is supplied.

In other words, what value is it supposed to use if it's not `true` or `false`? Like `null` or no value present at all? What then? So far as I can see... it blows up pretty badly.

NickGrantONS (dismissed):
Can't use a Java primitive when value might be null or not present at all in the JSON: guaranteed to cause problems

src/main/java/uk/gov/ons/census/casesvc/model/dto/Address.java
NickGrantONS: actually it might be ok...

NickGrantONS: okay, so the primitive thing does default to `false` but in that sense, we have no idea whether a value was supplied or not. This doesn't meet the requirements.

NickGrantONS (changes requested):
The requirement says, basically, "if the secure marker is supplied on the event, use it, otherwise copy the value from the source case".

That's not what's been implemented.

In order to know whether or not a value was supplied on the event, we'll have to use the `Optional` class. A dodgy hack we've used elsewhere is to use `Boolean` instead of `boolean` and assume that null means that no value was supplied, which is not strictly correct, but I'd accept that if you do it that way.

Then, once you know whether a value was supplier or **not** you can implement the business logic to use the value (true or false) if and when it was supplied, else take the value from the source case.

You can't use defaults for this feature, because it makes "false was supplied" and "we defaulted to false" indistinguishable, which doesn't allow for the value to be _not_ supplied, so the source case value is used.

Hope that's as clear as it can be.

src/main/java/uk/gov/ons/census/casesvc/service/NewAddressReportedService.java
```diff
@@ -288,7 +294,6 @@ private Case buildCaseFromSourceCaseAndEvent(CollectionCase newCollectionCase, C
     newCase.setOa(sourceCase.getOa());
     newCase.setPrintBatch(sourceCase.getPrintBatch());
     newCase.setSurvey(sourceCase.getSurvey());
-    newCase.setMetadata(metadataFromSourceCase(sourceCase.getMetadata()));
```
NickGrantONS, line 291: by removing this and using the `getMetadataFromEventOrSourceCase` we are potentially going to lose the `channel` metadata which would have been copied over - I see that as a bug

NickGrantONS: actually, I checked the original code, and it had the 'bug' too

NickGrantONS (approved):

ryangrundy7 (approved):
LGTM 👍

PR #210, opened 2020-09-29
Create or update after loading update file

gavinleeedwards:
# Motivation and Context
When dealing with address updates, if we sent a CREATE action to field for a case that they already know about then it causes them problems.  If an oa value is present for the original case then we need to send an UPDATE action instead.

# What has changed
If the oa is not null on the original case (before address updates have been applied) then we send an UPDATE action instead of a CREATE.

# How to test?
 - Run an address update (send a manual message or run the bulk updater from toolbox) for a case that doesn't have an oa value against it.  A CREATE field decision should be sent to field.
- Send another update against the same case. The case will now already have an oa value against it so an UPDATE should be sent to field.

# Links
- https://trello.com/c/vlpzyc7c

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
@@ -41,7 +41,7 @@ public class RmCaseUpdatedService {
 
     validateRmCaseUpdated(rmCaseUpdated);
 
-    boolean oaPresent = updatedCase.getOa() != null && !updatedCase.getOa().isEmpty();
+    boolean oaPresent = !StringUtils.isEmpty(updatedCase.getOa());
 
     updateCase(updatedCase, rmCaseUpdated);
 
@@ -56,9 +56,12 @@ public class RmCaseUpdatedService {
       eventMetadata = new Metadata();
       eventMetadata.setCauseEventType(rme.getEvent().getType());
 
-      // Only send a CREATE on a case that field doesn't already know about
-      // Disclaimer - we were forced by field to use this hack and it should be refactored
-      // properly when time allows.
+      // We don't want to send an UPDATE for cases which FWMT-G already know about,
+      // but the only way we've got of **guessing** that is by looking to see if an OA is on the
+      // case or not. We imagine that OA would only be set on cases which FWMT-G know about,
+      // so we have been forced to use it as an ugly kludge, because of time pressure.
+      // TODO: This should be refactored/done properly. It's tech debt.
+
       if (oaPresent) {
         eventMetadata.setFieldDecision(ActionInstructionType.UPDATE);
       } else {
--- a/src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
@@ -39,6 +39,8 @@ public class RmCaseUpdatedService {
     RmCaseUpdated rmCaseUpdated = rme.getPayload().getRmCaseUpdated();
     Case updatedCase = caseService.getCaseByCaseId(rmCaseUpdated.getCaseId());
 
+    boolean oaSet = updatedCase.getOa() != null;
+
     validateRmCaseUpdated(rmCaseUpdated);
 
     updateCase(updatedCase, rmCaseUpdated);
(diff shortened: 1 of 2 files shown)


gavinleeedwards:
[Create or Update after loading an update file (5)](https://trello.com/c/vlpzyc7c/1455-create-or-update-after-loading-an-update-file-5)

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
```diff
@@ -224,4 +232,8 @@ private void validateCase(Case caze) {
       throw new RuntimeException("Case missing mandatory fields after RM Case Updated");
     }
   }
+
+  public boolean isOaPresent(Case originalCase) {
+    return originalCase.getOa() != null;
+  }
```
ryangrundy7, line 238: This not used?

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
```diff
@@ -41,6 +41,8 @@ public void processMessage(ResponseManagementEvent rme, OffsetDateTime messageTi
 
     validateRmCaseUpdated(rmCaseUpdated);
 
+    boolean oaPresent = updatedCase.getOa() != null;
```
NickGrantONS, line 44: I probably would have preferred to use `StringUtils.isEmpty()` just in case we have a blank OA for some reason; somehow

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
```diff
@@ -53,7 +55,13 @@ public void processMessage(ResponseManagementEvent rme, OffsetDateTime messageTi
     if (shouldSendCaseToField(updatedCase, rme.getEvent().getChannel())) {
       eventMetadata = new Metadata();
       eventMetadata.setCauseEventType(rme.getEvent().getType());
-      eventMetadata.setFieldDecision(ActionInstructionType.CREATE);
+
+      // Only send a CREATE on a case that field doesn't already know about
```
NickGrantONS, line 59: For me, this "here be dragons" comment is not big and bold enough. Anybody reading our code will have absolutely no idea why the "OA being present" means "we should send an update" - we need to explain very clearly here, with big disclaimers, that we have been forced to do this, and we really don't want to be doing it like this - it's an ugly hack that we were forced to do; that it should be refactored away and done properly, when time allows; tech debt

ryangrundy7: ^ I think that should be the comment 😂

NickGrantONS (changes requested):
See comments

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
gavinleeedwards: Good spot.  That got left in by mistake.  Removed.

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
gavinleeedwards: The null check still needs to be there but I've added isEmpty() to the clause to cover blank strings that aren't null.

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
gavinleeedwards: I've updated the comment to add a disclaimer about us being forced to do it, should be refactored etc.  'Here be dragons' went right over my head 😂

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
```diff
@@ -41,6 +41,8 @@ public void processMessage(ResponseManagementEvent rme, OffsetDateTime messageTi
 
     validateRmCaseUpdated(rmCaseUpdated);
 
+    boolean oaPresent = updatedCase.getOa() != null && !updatedCase.getOa().isEmpty();
```
NickGrantONS, line 44:
```suggestion
    boolean oaPresent = !StringUtils.isEmpty(updatedCase.getOa());
```

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
NickGrantONS: I'd still like a little more explanation about what the heck this is.

Suggested wording:
```
HERE BE DRAGONS!!
We don't want to send an UPDATE for cases which FWMT-G already know about, but the only way we've got of **guessing** that is by looking to see if an OA is on the case or not. We imagine that OA would only be set on cases which FWMT-G know about, so we have been forced to use it as an ugly kludge, because of time pressure. TODO: This should be refactored/done properly. It's tech debt.
```

NickGrantONS (changes requested):
See minor comments

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
gavinleeedwards: Done 👍

src/main/java/uk/gov/ons/census/casesvc/service/RmCaseUpdatedService.java
gavinleeedwards: Done 👍

NickGrantONS (approved):

ryangrundy7 (approved):

PR #211, opened 2020-09-30
feature/rabbitmq-guest-user

apolloakora:
Security best practise urges us not to use default account usernames and passwords.

- **`rmquser`** - is the new standard rabbitmq user
- **`rmqmanager`** - is new metrics and monitoring user
- **`guest`** is to be deleted manually after release

Important : Do not forget when testing to give a BRANCH variable feature/rabbitmq-guest-user so that apply-kubernetes can download the correct tmp_rm_kube


apolloakora:
[Stop using default guest user for RabbitMQ (8)](https://trello.com/c/Ok5vnmxD/1440-stop-using-default-guest-user-for-rabbitmq-8)

PR #213, opened 2020-10-06
AddressTypeChange for CE Case ceActualResponses Defaulted to Zero

LeoKHoward:
# Motivation and Context
When an AddressTypeChange event occurs where the new address is a CE Case, the ceActualResponses is null rather than 0 

# What has changed
Changed ceActualResponses from Integer to int so now defaults to 0 instead of null
Test

# How to test?
Maven clean install on branch and check image builds and tests pass

# Links
https://trello.com/c/DUgevj8y

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
@@ -45,7 +45,9 @@ public class AddressTypeChangeService {
     if (oldCase.getCaseType().equals("HI")) {
       throw new RuntimeException("Cannot change case of type HI");
     }
-    addressModificationValidator.validate(addressTypeChange.getCollectionCase().getAddress());
+    addressModificationValidator.validate(
+        addressTypeChange.getCollectionCase().getAddress(),
+        responseManagementEvent.getEvent().getChannel());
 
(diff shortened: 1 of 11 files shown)


LeoKHoward:
[CE case missing ceActualResponses](https://trello.com/c/DUgevj8y/1494-ce-case-missing-ceactualresponses)

src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
```diff
@@ -98,6 +98,10 @@ private void createNewCase(
     newCase.setHtcWillingness(oldCase.getHtcWillingness());
     newCase.setHtcDigital(oldCase.getHtcDigital());
 
+    if (newCase.getCaseType().equals("CE")) {
```
NickGrantONS, line 101: Why only do this for CE cases? Are we taking the wrong approach completely. We could change `ceActualResponses` from an `Integer` to an `int` and save ourselves headache.

Could do the same with `ceExpectedCapacity` - make it an `int` and save ourselves a bunch of weird null pointer exceptions.

NickGrantONS (changes requested):
I think there's a better approach. See comments.

NickGrantONS (approved):
Looks good. I wonder if we're going to get null pointer exceptions when people start sending us new cases with null `ceActualResponses` but that's their problem, not ours

ryangrundy7 (approved):

PR #214, opened 2020-10-13
Allow Contact Centre to send Estab Type OTHER for modifications and case type change modifications

NickGrantONS:
# Motivation and Context
I didn't want to do this, but it was too hard for other systems to make the necessary changes, so we had to kludge RM. I don't know. Read the Trello ticket for some background.

# What has changed
Allowed Contact Centre to send an estab type of "OTHER" although it is not one of the allowed ones.

# How to test?
Try sending an address modification with estab type "OTHER" from CC, and it should work. Try sending an address modification with estab type "OTHER" from RH, and it should throw an exception.

Try sending an address type change, with an address modification, with estab type "OTHER" from CC, and it should work. Try sending an an address type change, with an address modification, with estab type "OTHER" from RH, and it should throw an exception.

# Links
Trello: https://trello.com/c/ZUuPMYuQ

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/AddressModificationService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/AddressModificationService.java
@@ -33,7 +33,8 @@ public class AddressModificationService {
 
     Case caze = caseService.getCaseByCaseId(addressModification.getCollectionCase().getId());
 
-    addressModificationValidator.validate(addressModification.getNewAddress());
+    addressModificationValidator.validate(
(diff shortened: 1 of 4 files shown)


NickGrantONS:
[Allow "OTHER" estab type in address modified events from Contact Centre, also via address type changed including modification](https://trello.com/c/ZUuPMYuQ/1604-allow-other-estab-type-in-address-modified-events-from-contact-centre-also-via-address-type-changed-including-modification)

hbrace (approved):
Works for me 👍

LukeLoze1 (approved):

PR #216, opened 2020-11-02
Send UPDATE to FWMT-G for FieldCaseUpdated (CE Units only)

NickGrantONS:
# Motivation and Context
Business requirements have changed.

# What has changed
Only send to field CE units. Send update if the actual response count is greater than the expected response count.

# How to test?
Try updating the case using FIELD_CASE_UPDATED event on CE unit cases. When the actual response count is equal to or greater than the expected response count, the Field case should be cancelled, otherwise it should be updated. Other case types and address levels should either be ignored or throw errors, per prior behaviour.

# Links
Trello: https://trello.com/c/E1Qk7n8H


NickGrantONS:
[Change request: send updates to field when expected response count changed (5)](https://trello.com/c/E1Qk7n8H/1639-defect-sending-cancel-instead-of-update-when-receiving-a-field-case-updated-event)

ryangrundy7 (approved):

hbrace (approved):

PR #217, opened 2020-11-02
Redact sensitive data

NickGrantONS:
# Motivation and Context
We have been outputting UACs into our event log, which means they're visible via Case API, and therefore the ROps UI too. Generally speaking, we should never leak UACs because they allow access to respondents confidential census answers. Although this would be systems internal and all staff should have the right clearance, it's better practice to treat this data on a need to know basis, and keep it hidden.

# What has changed
Redact away UACs before writing event payloads to the DB.

# How to test?
Do some events which would have logged events. Now the UACs should be redacted.

# Links
Trello: https://trello.com/c/XYihYjv7


NickGrantONS:
[Strip out UAC from event log payload for telephone capture fulfilments (8)](https://trello.com/c/XYihYjv7/1626-strip-out-uac-from-event-log-payload-for-telephone-capture-fulfilments-8)

src/main/java/uk/gov/ons/census/casesvc/utility/RedactHelper.java
```diff
@@ -0,0 +1,57 @@
...
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+
+public class RedactHelper {
+  private static final String REDACTION_FAILURE = "Failed to redact sensitive data";
+  private static final String REDACTION_TARGET = "setUac";
+  private static final String REDACT_TO_THIS_VALUE = "REDACTED";
+  private static final ObjectMapper objectMapper = ObjectMapperFactory.objectMapper();
+
+  public static Object redact(Object rootObjectToRedact) {
```
AdamHawtin, line 15: Could be named explicitly `redactUacs`

src/main/java/uk/gov/ons/census/casesvc/utility/RedactHelper.java
```diff
@@ -0,0 +1,57 @@
...
+    Arrays.stream(object.getClass().getMethods())
+        .filter(item -> Modifier.isPublic(item.getModifiers()))
+        .forEach(
+            method -> {
+              if (method.getName().startsWith("get")
+                  && method.getReturnType().getPackageName().equals(packageName)) {
+                try {
+                  Object invokeResult = method.invoke(object);
+                  if (invokeResult != null) {
+                    recursivelyRedact(invokeResult, methodTofind, packageName);
```
AdamHawtin, line 40: Is there a danger of infinite recursion if two classes reference each other?

AdamHawtin (approved):
A couple of minor comments but nothing blocking 👍

ryangrundy7 (approved):
Looks good! Looked in the DB, case api and r ops and all uacs I could see were redacted

src/main/java/uk/gov/ons/census/casesvc/utility/RedactHelper.java
NickGrantONS: deliberately not in case we want this to redact more than just UACs

src/main/java/uk/gov/ons/census/casesvc/utility/RedactHelper.java
NickGrantONS: yes it will do that, but that would be caught in unit or integration testing. I haven't tried to write a helper class for use in any and all Comp Sci scenarios, but instead a redactor for use on this project

PR #220, opened 2020-11-06
Remove sending UPDATE to field when ER > AR

NickGrantONS:
# Motivation and Context
When the ticket was implemented, it said to send an UPDATE to field when ER > AR. The ticket was then amended, by whom and why is not known. The desired behaviour from product manager is _currently_ at time of writing, to only send CANCEL and never send UPDATE.

# What has changed
Removed UPDATE behaviour.

# How to test?
Don't bother. It's going to go through PO review anyway.

# Links
Trello: https://trello.com/c/C7cvRnRv


NickGrantONS:
[Don't send updates on field case updated](https://trello.com/c/C7cvRnRv/1678-dont-send-updates-on-field-case-updated)

hbrace (approved):

liamtoozer (approved):

PR #221, opened 2020-11-13
CCS Property Listed Shouldn't Use Default Rules

NickGrantONS:
# Motivation and Context
The default "don't send to field if..." rules are stopping CCS cases from going to field when they're missing certain data. Apparently we don't want that.

# What has changed
Removed the default rule check from CCS property listed.

# How to test?
Run the new unit test.

# Links
Trello: https://trello.com/c/YDlvY9oL


NickGrantONS:
[Do not use default send to field rules for CCS cases. (3)](https://trello.com/c/YDlvY9oL/1692-do-not-use-default-send-to-field-rules-for-ccs-cases-3)

AdamHawtin (approved):
👍 Does what it says on the tin

davidmort (approved):

PR #222, opened 2020-11-13
isHouseholder Flag Added To Refusal Event

LeoKHoward:
# Motivation and Context
Need to know if the respondent who refused is the householder along with their name and address details
isHouseholder flag is independent of Contact and Address blocks in event JSON so even if isHouseholder is false, Contact and Address may still have details in them

# What has changed
isHouseholder flag added to Refusal DTO (defaults to false if not supplied)
Contact and Address fields added to Refusal DTO
Report removed
Tests added

# How to test?
Maven Clean Install
Check tests pass and image builds
Run ATs with new image (although changes are minimal)
Try sending refusal event JSON on rabbit to check that isHouseholder is true when set and is false otherwise (especially when not included on event)

# Links
https://trello.com/c/WT0EGYyU/1150-nc-update-refusal-event-5
https://collaborate2.ons.gov.uk/confluence/display/CATD/Response+Management+Event+Dictionary?focusedCommentId=20210501#comment-20210501

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
@@ -14,7 +14,8 @@ public class RefusalDTO {
   private String callId;
   private boolean isHouseholder;
 
-//  need to define getter here or else isHouseholder flag will default to householder in event JSON and it wont work
+  //  need to define getter here or else isHouseholder flag will default to householder in event
+  // JSON and it wont work
   public boolean getIsHouseholder() {
     return this.isHouseholder;
   }
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
@@ -2,6 +2,7 @@ package uk.gov.ons.census.casesvc.model.dto;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.databind.JsonNode;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
@@ -10,8 +11,12 @@ import lombok.NoArgsConstructor;
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private boolean isHouseholder;
   private CollectionCase collectionCase;
+
+  private JsonNode contact;
+  private JsonNode address;
+
 }
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
@@ -12,7 +12,7 @@ public class RefusalDTO {
   private RefusalTypeDTO type;
   private String agentId;
   private String callId;
-  private boolean isHouseholder;
+  private Boolean isHouseholder;
   private CollectionCase collectionCase;
   private Contact contact;
   private Address address;
--- a/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
@@ -2,7 +2,6 @@ package uk.gov.ons.census.casesvc.model.dto;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.databind.JsonNode;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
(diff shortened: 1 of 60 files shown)


LeoKHoward:
[NC - update refusal event (5)](https://trello.com/c/WT0EGYyU/1150-nc-update-refusal-event-5)

src/main/java/uk/gov/ons/census/casesvc/service/RefusalService.java
```diff
@@ -36,6 +36,10 @@ public void processRefusal(
       return;
     }
 
+    if (refusalDto.getIsHouseholder() == null) {
```
NickGrantONS, line 39: if you're going to default it to false, you can just used a primitive instead of a class

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +10,10 @@
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private Boolean isHouseholder;
   private CollectionCase collectionCase;
+  private Contact contact;
```
NickGrantONS, line 17: I'm not sure about this. I thought this was supposed to be an encrypted blob

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +10,10 @@
...
+  private Address address;
```
NickGrantONS, line 18: Is this supposed to be encrypted? Possibly not, but still...

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
NickGrantONS: okay I checked the event dictionary. this looks OK

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
NickGrantONS: okay I checked the event dictionary. this looks OK

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +10,10 @@
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private Boolean isHouseholder;
```
NickGrantONS, line 15: prefer primitives everywhere, except where null is explicitly wanted, which is not the case here

NickGrantONS (changes requested):
need to use primitives, unless null is meaningful

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +11,16 @@
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private boolean isHouseholder;
+
+  @JsonProperty("isHouseholder")
```
NickGrantONS, line 18: this is not necessary

NickGrantONS: There's no need to use the `@JsonProperty` annotation - you can remove it. It's not doing anything useful

NickGrantONS (changes requested):
See comment

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
LeoKHoward: It sets the flag in the DB to the name isHouseholder rather than changing it to householder. If I remove it, the flag will always be false even if it is set to true in the event message. With it, it makes the difference.

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +10,15 @@
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private boolean isHouseholder;
+
+  public boolean getIsHouseholder() {
```
NickGrantONS, line 17: Can you put a comment here to say that this is needed otherwise we will get `householder` in the JSON instead of `isHouseholder` pls

NickGrantONS (approved):
Minor request for a comment to clarify why ugly code hack needed.

src/main/java/uk/gov/ons/census/casesvc/model/dto/RefusalDTO.java
```diff
@@ -10,8 +10,17 @@
 @JsonInclude(Include.NON_NULL)
 public class RefusalDTO {
   private RefusalTypeDTO type;
-  private String report;
   private String agentId;
   private String callId;
+  private boolean isHouseholder;
+
+  //  need to define getter here or else isHouseholder flag will default to householder in event
+  // JSON and it wont work
```
NickGrantONS, line 18: 👍

LukeLoze1 (approved):

PR #230, opened 2021-03-09
Check if indiv case has already beeen created

NickGrantONS:
# Motivation and Context
There's a bug when we process the same individual fulfilment message more than once, then Hibernate converts the INSERT of new case into an UPDATE, and mangles the data (deletes created timestamp and sets secret sequence number to zero, and then of course, sets case ref to be the same one every single time).

# What has changed
Check if individual case ID already exists; throw exception if it does.

# How to test?
Send an individiual fulfilment. Check that the case has been created. Send the same message AGAIN. This should result in "mangled" case data as described.

# Links
Trello: https://trello.com/c/cH8Gw1x2

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
@@ -115,7 +115,7 @@ public class FulfilmentRequestService {
         throw new RuntimeException("Individual fulfilment must have an individual case ID");
       }
 
-      if (caseService.getCaseByCaseId(fulfilmentRequestPayload.getIndividualCaseId()) != null) {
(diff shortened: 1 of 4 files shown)


LukeLoze1 (approved):

src/main/java/uk/gov/ons/census/casesvc/service/FulfilmentRequestService.java
```diff
@@ -115,6 +115,10 @@ private void handleIndividualFulfilmentForHHCase(
         throw new RuntimeException("Individual fulfilment must have an individual case ID");
       }
 
+      if (caseService.getCaseByCaseId(fulfilmentRequestPayload.getIndividualCaseId()) != null) {
+        throw new RuntimeException("Duplicate message");
```
AdamHawtin, line 119: Could provide a more specific exception message

AdamHawtin (approved):

ryangrundy7 (approved):

PR #231, opened 2021-03-10
HOTFIX: Fix other places where dupe message might mess up cases

NickGrantONS:
# Motivation and Context
We have a known issue when we are provided with a supposedly new and unique case ID in a message, but the message gets delivered: in this instance, it's likely that we will end up creating a second case with a secretSequenceNumber of zero, and a case ref which has already been used. This behaviour is an annoying aspect of Hibernate trying to be helpful when you ask it to save an entity to the database, and you provide the ID (primary key) of an _existing_ row.

# What has changed
Fixed all the places where this kind of bug might occur (although very unlikely to happen).

# How to test?
Run the ATs. Zero regression.

# Links
Trello: https://trello.com/c/6XXXsrvm

### DIFF
--- a/src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
+++ b/src/main/java/uk/gov/ons/census/casesvc/service/AddressTypeChangeService.java
@@ -76,6 +76,11 @@ public class AddressTypeChangeService {
       OffsetDateTime messageTimestamp,
       AddressTypeChange addressTypeChange,
       Case oldCase) {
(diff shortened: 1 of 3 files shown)


NickGrantONS:
[HOTFIX: handling duplicate message delivery when creating new cases](https://trello.com/c/6XXXsrvm/2119-hotfix-handling-duplicate-message-delivery-when-creating-new-cases)

ryangrundy7 (approved):

AdamHawtin (approved):
I would like some tests for these paths but happy to backfill them later to get this in quick now, since we already have a careful plan to test manually.

